PR #8, opened 2019-01-09
logging to console (behind flag)

### LINKED ISSUE #7: Logging for steps
Init and processing of steps should be logged to the console in verbose mode.

tpluscode:
Fixes #7 

@bergos, I added a simple stream which pushes log strings for events with a name. That should be enough for us for now.

Do you have anything else in mind?


bin/barnard59.js
```diff
@@ -63,7 +64,8 @@ program
   .option('--output [filename]', 'output file', '-')
   .option('--pipeline [iri]', 'IRI of the pipeline description')
   .option('--variable <name=value>', 'variable key value pairs separated by comma', parseVariables)
-  .action((filename, { format, output, pipeline, variable } = {}) => {
+  .option('--verbose', 'enable diagnostic console output')
```
bergos: Can you make it `'-v, --verbose'`? I think most cli also support `-v`, so people may try to use the short version.

bergos (commented):
Can you move `bin/TextLogStream.js` to ` lib/TextLogStream.js`? There are no clear rules for the folder structure, but I think usually only the code which is used in the `bin` property of `package.json` is in the `bin` folder.

bergos (approved):

PR #10, opened 2019-01-09
added shell support and examples

bergos:
This PR adds the `barnard59-shell` dependency and some examples which show how to use the shell operations.


tpluscode (approved):
For now I think it's good. We should continue the discussion around before/after feature but otherwise I think we should not be held back by that talks.

I'd maybe just add a section about the shell example to the readme similar to the other two paragraphs.

PR #17, opened 2019-04-23
added concept and pipeline definition docs

bergos:
This PR adds documentation about the basic concept, internals (stream stuff), pipeline definitions (steps, variables) and how JS code can be used in the pipeline.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,213 @@
 # barnard59
 
+`barnard59` runs pipelines which are described in RDF in the [pipeline](https://pipeline.described.at/) and [code](https://code.described.at/) ontology.
+
+## Concept
+
+A barnard59 pipeline consists of one or more steps where each step returns a [Stream](https://nodejs.org/api/stream.html).
+The streams of all steps are combined via [.pipe](https://nodejs.org/api/stream.html#stream_readable_pipe_destination_options).
+The pipeline itself is also a stream.
+Depending on the stream types of the first and last stream, the pipeline can be writable or readable.
+If the pipeline is self-contained, a dummy readable interface is provided to handle events for end of processing and errors.
+
+### Internals
+
+The code to process pipelines makes heavy usage of [RDF Loaders](https://github.com/zazuko/rdf-native-loader) to create JavaScript objects and functions based on a RDF description.
+For the steps, the [Code Ontology](https://code.described.at/) is used to point or define the code for each step.
+The [RDF Code Loader](https://github.com/zazuko/rdf-native-loader-code) takes care of importing code referenced as IRIs or directly process literals as JavaScript code.  
+
+## Pipeline Definition
+
+The following prefixes are used in the code segments of this section:
+
+```turtle
+@prefix code: <https://code.described.at/> .
+@prefix p: <https://pipeline.described.at/> .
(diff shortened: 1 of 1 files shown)


vhf (approved):

README.md
```diff
@@ -1,5 +1,213 @@
 # barnard59
 
+`barnard59` runs pipelines which are described in RDF in the [pipeline](https://pipeline.described.at/) and [code](https://code.described.at/) ontology.
+
+## Concept
+
+A barnard59 pipeline consists of one or more steps where each step returns a [Stream](https://nodejs.org/api/stream.html).
+The streams of all steps are combined via [.pipe](https://nodejs.org/api/stream.html#stream_readable_pipe_destination_options).
```
tpluscode:
```suggestion
The streams of all steps are combined via [`.pipe`](https://nodejs.org/api/stream.html#stream_readable_pipe_destination_options).
```

README.md
```diff
@@ -1,5 +1,213 @@
...
+<#pipeline> a p:Pipeline, p:Writable, p:Readable .
+```
+
+**Note:** The type of the streams could be also detected during runtime or based on metadata for the streams, but that would required deeper understanding of the pipeline and steps structures.
+To simplify handling of pipeline definitions without runtime data, the stream types are attached to the pipeline definition and errors are thrown in the runtime if they don't match. 
+
+### Step
+
+A pipeline consists of one or more steps, which are actually factories which return streams.
+Each step is linked with the previous one via [.pipe](https://nodejs.org/api/stream.html#stream_readable_pipe_destination_options).
```
tpluscode:
```suggestion
Each step is linked with the previous one via [`.pipe`](https://nodejs.org/api/stream.html#stream_readable_pipe_destination_options).
```

README.md
```diff
@@ -1,5 +1,213 @@
...
+  code:arguments [
+    p:name "url";
+    p:value "http://example.org/"
+  ], [
+    p:name "method";
+    p:value "POST"
+  ] .
+```
+
+The function in `customStreamCode.js` will be called with the following object:
```
tpluscode: I would make it clear somewhere in this section that the implementation must export function, which takes loaded `code:arguments` and returns a `Stream`.

README.md
```diff
@@ -1,5 +1,213 @@
...
+  code:implementedBy [ a code:EcmaScript ;
+    code:link <node:barnard59-base#map>
+  ] ;
+  code:arguments ("""row => {
+    // replace all ö umlauts with oe
+    return row.split('ö').join('oe')
+  }"""^^code:EcmaScript) .
+```
+
+#### EcmaScript Template Literal
```
tpluscode, line 199: This section kind of repeats `Using Variables in Template Literals`

tpluscode (changes requested):

README.md
bergos: I know. Made the structure, then filled it. Somehow it should be there as separate topic. Maybe we fill it in the future with more stuff how the template literals can be useful and point from the variable section to this one? Maybe even reduce the stuff in the variable section?

tpluscode: m'kay. let's keep it for now.

tpluscode (approved):

PR #33, opened 2020-01-24
refactor: extract the runner function from bin

tpluscode:
- [x] refactoring
- [x] type declaration for new code
- [ ] mention in readme

I want to be able to trigger a pipeline programmatically rather than calling `barnard59` from command line

For that, I extracted the core functionality from `bin/barnard59.js`, removing its dependencies on module `fs` and `process`.

Usage will be similar to what the bin does:

```js
const { create } = require('barnard59/lib/runner')

const run = create({
  pipeline: 'http://example.com/MyPipeline', 
  variable: [],
  outputStream: process.stdout,
  basePath: __dirname
})

const dataset = myInMemoryPipelineTriples()

run(dataset)
```


lib/runner.js
```diff
@@ -0,0 +1,83 @@
...
+
+    return arr
+  }, [])
+
+  if (rootPipelines.length > 1) {
+    throw new Error('multiple root pipeline found. please specify the one to run using --pipeline option')
+  }
+
+  return rootPipelines[0]
+}
```
bergos: I think guessing the right pipeline should stay in the cli code (or some util function) and in the API the `pipeline` argument should be required. That removes some `magic` from the logic and makes it more understandable and easier to document.

lib/runner.js
```diff
@@ -0,0 +1,83 @@
...
+
+function create ({ basePath, outputStream, pipeline, variable, verbose, log } = {}) {
+  log = log || function () { }
```
bergos: I think this will cause a problem in...

lib/runner.js
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    stream.on('error', err => {
+      throw err
+    })
+
+    stream.pipe(outputStream)
+
+    if (verbose) {
+      stream.context.log.pipe(new TextLogStream()).pipe(log)
```
bergos: ...this line, cause it expects a stream. You can change it to `.on('data', text => log(text))`.

Maybe instead of using the `verbose` argument and `log()`, always call [debug](https://www.npmjs.com/package/debug). The cli tool could programmatically enabled debug for this namespace.

lib/runner.d.ts
```diff
@@ -0,0 +1,19 @@
...
+    outputStream: Writable
+    pipeline?: string
+    variable?: Map<string, string>
+    verbose?: boolean
+    log?: (message: string) => void
+}
+
+declare function create (options: RunnerInit): Runner
+
+export = create
```
bergos: Don't know what happened here, but doesn't look like it should be part of the PR.

bergos (changes requested):

lib/runner.d.ts
tpluscode: I did mention that in the PR description

lib/runner.js
tpluscode: Oh shoot, a stream. And yea, the suggestion to always go for debug makes total sense

lib/runner.js
```diff
@@ -0,0 +1,43 @@
+const p = require('..')
+const TextLogStream = require('../lib/TextLogStream')
+
+const log = p.log.extend('runner')
+log.enabled = false
```
tpluscode, line 5: That's the best I came up with. A runner logger will use namespace `barnard59:runner`, where the first segment is imported from the main module.

Disabled by default.

bergos (approved):
Looks good.

PR #37, opened 2020-02-14
refactor: simplify the calls to use runner programmatically

tpluscode:


bergos:
The next branch provides an interface with a graph pointer argument and makes this PR deprecated.

PR #38, opened 2020-03-09
added concat example and updated dep

bergos:


martinmaillard (approved):

ktk (approved):
LGTM. This has to be done on Pipeline or can it also be used to concat two steps?

PR #41, opened 2020-04-01
added buffer debug, changed runner interface

bergos:
- adds an option to show a histogram of the buffer usage that can be used to detect back pressure problems
- changed the runner interface to have access to the finished promise + the pipeline

### DIFF
--- a/bin/barnard59.js
+++ b/bin/barnard59.js
@@ -7,6 +7,7 @@ const program = require('commander')
 const cf = require('clownface')
 const rdf = require('rdf-ext')
 const namespace = require('@rdfjs/namespace')
+const bufferDebug = require('../lib/bufferDebug')
 const runner = require('../lib/runner')
 
 const ns = {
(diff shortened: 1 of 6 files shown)


bin/barnard59.js
```diff
@@ -65,30 +66,36 @@ program
...
   .option('-v, --verbose', 'enable diagnostic console output')
-  .action((filename, options = {}) => {
-    let { format, output, pipeline, verbose } = options
+  .option('--debug-buffer', 'enable histogram of buffer usage')
+  .action(async (filename, options = {}) => {
+    try {
+      let { format, output, pipeline, verbose, debugBuffer } = options
 
-    runner.log.enabled = verbose
+      runner.log.enabled = verbose
```
tpluscode, line 74:
```suggestion
      runner.log.enabled = verbose && !debugBuffer
```

bergos: I know what you mean. Let's see if there are any problems running both at the same time. If there are problems, let's see if we can fix it or throw already an error if both options are given at the same time.

tpluscode: Fair enough

tpluscode (approved):

PR #45, opened 2021-01-27
fix: upgraded barnard-* packages

bergos:


package.json
```diff
@@ -19,8 +19,8 @@
   "homepage": "https://github.com/zazuko/barnard59",
   "dependencies": {
     "@rdfjs/namespace": "^1.0.0",
-    "barnard59-base": "^0.0.4",
-    "barnard59-core": "^0.0.6",
+    "barnard59-base": "0.0.5",
```
tpluscode: I propose we merge https://github.com/zazuko/barnard59-base/pull/12 and publish `barnard59-base: 0.1`

In general, I would like to get rid of `0.0.x` packages because they are treated as incompatible with any other

bergos: @tpluscode makes sense. Did that for `-core`, wasn't aware of the open PR in `-base`. Just merged the PR, released it as `0.1.0` and upgraded the dependency in this package.

tpluscode (approved):

PR #56, opened 2021-09-01
OpenTelemetry improvements

tpluscode:
1. Added metrics exporter
2. Do not enable console diag output by default
3. Update `@opentelemetry` packages


bin/barnard59.js
```diff
@@ -11,8 +11,6 @@ import { BatchSpanProcessor } from '@opentelemetry/tracing'
 
 import { Option, Command } from 'commander'
 
-diag.setLogger(new DiagConsoleLogger())
-
```
tpluscode, line 15: Make optional because it tends to flood the standard output with warning message `Dropping extra events`

bin/barnard59.js
```diff
@@ -64,6 +67,19 @@ const onError = async err => {
...
   }
+  // Export the metrics to a collector. By default it exports to
+  // http://localhost:55681/v1/metrics, but it can be changed with the
+  // OTEL_EXPORTER_OTLP_METRICS_ENDPOINT environment variable.
+  if (otelMetricsExporter === 'otlp') {
+    const exporter = new CollectorMetricExporter()
+    sdk.configureMeterProvider({
+      exporter
+    })
+  }
```
tpluscode: Similar to `otelExporterOpt` above

bin/barnard59.js
```diff
@@ -64,6 +67,19 @@ const onError = async err => {
     const spanProcessor = new BatchSpanProcessor(exporter)
     sdk.configureTracerProvider({}, spanProcessor)
   }
+  // Export the metrics to a collector. By default it exports to
+  // http://localhost:55681/v1/metrics, but it can be changed with the
+  // OTEL_EXPORTER_OTLP_METRICS_ENDPOINT environment variable.
+  if (otelMetricsExporter === 'otlp') {
+    const exporter = new CollectorMetricExporter()
+    sdk.configureMeterProvider({
+      exporter
```
sandhose:
```suggestion
      exporter,
      interval: 10000, // Export metrics every 10 seconds
```

bin/barnard59.js
sandhose: I guess you could also do `diag.setLogger(new DiagConsoleLogger(), DiagLogLevel.ERROR);` instead, that particular log line is at the `warn` level

sandhose (approved):
LGTM overall. Could you also bump the OTEL libraries to 0.24?

bin/barnard59.js
tpluscode: Oh, I would not like to hardcode that. Override with env var maybe?

bin/barnard59.js
tpluscode: So what about making this `--otel-debug <level>` with values matching the enum?

bin/barnard59.js
tpluscode: Or an additional `--otel-metrics-interval <ms>` option

sandhose (approved):
Looks pretty good to me!

bergos (approved):

PR #67, opened 2023-01-03
feat: hiding sensitive variables

### LINKED ISSUE #66: Hide variables from output
With the verbose `-v` flag, the CLI prints info level messages to the standard output (error stream? unimportant)

That may not be ideal in case of sensitive values, such as passwords. Maybe we could have a way to annotate such variables so that the are not exposed in the output? For example:

```turtle
<Main>
  a :Pipeline, :ReadableObjectMode ;
  :variables
    [
      :variable
        [
          a :Variable ;
          :name "API_PASSWORD" ;
          :sensitive true ;
        ] ;
    ] ;
.
```

Then they would be replaced with a bunch of asterisks

```
info: variables via runner:
info:   API_PASSWORD: ***
```

tpluscode:
Proposal to fix #66

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,7 +51,9 @@
     "mocha": "^10.2.0",
     "shelljs": "^0.8.4",
     "stricter-standard": "^0.2.0",
-    "strip-ansi": "^7.0.0"
+    "strip-ansi": "^7.0.0",
+    "winston": "^3.3",
+    "winston-transport": "^4.4.0"
   },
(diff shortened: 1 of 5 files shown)


tpluscode:
Did I cause the failing test? I don't see how my changes would have affected it...

tpluscode:
Closing in order to move variable logging to `barnard59-core`

PR #84, opened 2023-04-05
chore: set up changesets

tpluscode:


.github/workflows/release.yaml
```diff
@@ -0,0 +1,33 @@
...
+    steps:
+      - name: Checkout Repo
+        uses: actions/checkout@v3
+
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 18
+
+      - name: Install Dependencies
+        run: yarn
```
ludovicm67: I see a `package-lock.json` file, I guess it will be better if we use the `npm ci` command instead of `yarn` here, no?

```suggestion
        run: npm ci
```

ludovicm67 (approved):

.github/workflows/release.yaml
tpluscode: Very much so :)

PR #89, opened 2023-04-17
build(deps): update packages

tpluscode:


test/barnard59.test.js
```diff
@@ -112,9 +112,9 @@ describe('barnard59', () => {
 
     it('should run the parse-csvw.ttl example without error', () => {
       const pipelineFile = (new URL('../examples/parse-csvw.ttl', import.meta.url)).pathname
-      const command = `${barnard59} run --pipeline=http://example.org/pipeline/parseCsvw ${pipelineFile}`
+      const command = `${barnard59} run -vv --pipeline=http://example.org/pipeline/parseCsvw ${pipelineFile}`
 
-      const result = shell.exec(command, { silent: true })
+      const result = shell.exec(command)
```
ludovicm67: Regarding this test, is it something that should be changed, or is it just a change that was added to this commit by mistake?

tpluscode: Indeed, thanks

ludovicm67 (approved):
As tests are passing, this looks good for me!

Upgrading `rdf-ext` to v2 will not have any impact on any pipeline? (just asking because of the `patch` level upgrade that needs to not break anything for existing pipelines).

tpluscode:
> Upgrading `rdf-ext` to v2 will not have any impact on any pipeline?

Short answer: it should not

Long answer: it should not as long as consumers use the base RDF/JS interface for datasets. For example, `rdf-parser-csvw` used the `toArray` method which is `rdf-ext`-only even though it did not even have a dependency on rdf-ext. In such case a pipeline may fail but I'd consider this a bug in downstream.

tpluscode:
The mentioned issue in `rdf-parser-csvw` has been fixed already

ludovicm67:
Ok, great! Thanks for the feedback here 👍 
LGTM.

Issue #98, opened 2023-05-09
Create a "barnard59-curl" package


tpluscode:
As proposed by @ktk, the first 100% reusable pipeline could be a self-contained replacement for curl.

Inputs: 
- a glob of local files to upload
- SPARQL endpoint details
- (optional) HTTP method

Output: none - the pipeline will use the graph store to upload

Might consider an option to silently continue (with a warning logged) when a file is not RDF, such as if a globbed directory contains a `readme.md` etc

ktk:
Basically a version of this that accepts [globs](https://www.npmjs.com/package/glob) & chooses the correct parser per filetype:

```turtle
@base         <urn:pipeline:xrm> .
@prefix :     <https://pipeline.described.at/> .
@prefix code: <https://code.described.at/> .
@prefix rdf:  <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .

<#defaultVars>
  :variable [ a :Variable;
    :name "outputFile";
    :value "target/all.nq"
  ]
  .

<#PostToStore> a :Pipeline;
  :variables <#defaultVars>;
  :steps [
    :stepList (
      <#readFile>
      <#parseN3>
      <#setGraph>
      <#postToGraphStore>
    )
  ].

<#PutToStore> a :Pipeline;
  :variables <#defaultVars>;
  :steps [
    :stepList (
      <#readFile>
      <#parseN3>
      <#setGraph>
      <#putToGraphStore>
    )
  ].

<#readFile> a :Step;
  code:implementedBy [ a code:EcmaScript;
    code:link <node:fs#createReadStream>
  ];
  code:arguments ("outputFile"^^:VariableName).


<#parseN3> a :Step;
  code:implementedBy [ a code:EcmaScriptModule;
    code:link <node:barnard59-formats/n3.js#parse>
  ].

<#putToGraphStore>
  a                   :Step;
  code:implementedBy  [ a         code:EcmaScriptModule;
                        code:link <node:barnard59-graph-store/put.js#default> ] ;
  code:arguments      [ code:name   "endpoint" ;
                        code:value  "ENDPOINT"^^:VariableName ] ,
                      [ code:name   "user" ;
                        code:value  "USER"^^:VariableName ] ,
                      [ code:name   "password";
                        code:value  "PASSWORD"^^:VariableName ].


<#postToGraphStore>
  a                   :Step;
  code:implementedBy  [ a         code:EcmaScriptModule;
[...]

PR #100, opened 2023-06-16
feat: fs/parse step

tpluscode:
I added the step mentioned in https://github.com/zazuko/barnard59/issues/98#issuecomment-1552819728

This will save us the hassle of setting up in a pipeline the steps to read and parse the file. Not only fewer steps, but also automatically chooses the right parser based on the file extension

### DIFF
--- a/packages/rdf/manifest.ttl
+++ b/packages/rdf/manifest.ttl
@@ -44,3 +44,14 @@
  code:implementedBy [ a code:EcmaScript;
                       code:link <node:barnard59-rdf/metadata.js#voidStats>
                     ].
+
+<fs.js#parse>
+  a p:Operation, p:ReadableObjectMode ;
+  rdfs:label "Parse RDF file" ;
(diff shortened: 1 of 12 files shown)


packages/rdf/manifest.ttl
```diff
@@ -44,3 +44,14 @@
  code:implementedBy [ a code:EcmaScript;
                       code:link <node:barnard59-rdf/metadata.js#voidStats>
                     ].
+
+<fs.js#parse>
+  a p:Operation, p:ReadableObjectMode ;
+  rdfs:label "Parse RDF file" ;
+  rdfs:comment "Opens and parses an RDF file, choosing the correct parser based on the extension" ;
```
giacomociti, line 51: is it worth listing the supported extensions? not sure because this depends on an external library which may evolve independently

giacomociti (approved):

packages/rdf/manifest.ttl
tpluscode: Good idea, but I will put it in the readme instead

Issue #31, opened 2019-11-25
Load files from the local file system as arguments


bergos:
In some cases it would be useful to read the content of files on the local file system. For example a ssh key can be loaded like this:

```
"${require('fs').readFileSync('/full/path/myKeyFile.key').toString()}"^^code:EcmaScriptTemplateLiteral
```

But it would be useful to have a datatype + loader that handles it:

```
"/full/path/myKeyFile.key"^^p:FileContent
```

It should also work with variables:

```
"keyFilename"^^p:VariableFileContent
```

tpluscode:
This could also be implemented as as stream itself. The SHACL validation step does that to [load the shapes graph](https://github.com/zazuko/barnard59-validate-shacl/blob/a5ecc1b0f843d5de17355935eed74a11c546abea/validate.js#L55). That is not as nice as a synchronous loader to plug in directly to the function but avoids the risk of loading large files to memory

The problem of using a pipeline is verbosity, requiring a more complex setup. Given the original example 

```turtle
<step>
  a :Step ;
  code:arguments (
    <loadStepFile>
  ) ;
.

<loadStepFile> a :Pipeline, :Readable ;
  :steps [ :stepList (
    [ 
      code:implementedBy [ code:link <node:fs#createReadStream> ] ;
      code:arguments ( "/full/path/myKeyFile.key" )
    ] 
  ) ] ;
.
```

Hypothetically, #85 could be a solution, where the `<loadStepFile>` becomes a reusable building block

PR #142, opened 2023-07-31
File loader for `rdf-loaders-registry`

tpluscode:
I added a new loader which synchronously reads the contents of a file so that is can be easily injected as step argument

The difference from the original proposal in #31 is that I do not differentiate between path being used directly and path as variable. `"/full/path/myKeyFile.key"^^p:FileContent` and `"KEY_FILE"^^p:FileContent` are going to have the same effect as long as there is a variable called `KEY_FILE` in the pipeline's context


ludovicm67:
Great, thanks for implementing this feature @tpluscode!

Regarding this:

> The difference from the original proposal in https://github.com/zazuko/barnard59/issues/31 is that I do not differentiate between path being used directly and path as variable. "/full/path/myKeyFile.key"^^p:FileContent and "KEY_FILE"^^p:FileContent are going to have the same effect as long as there is a variable called KEY_FILE in the pipeline's context

This might be a bit confusing, since it will have to deal with the two options, and we need to make sure we won't have any conflict there (should be mostly OK, since a path looks different enough than a variable name), and make sure that it is properly documented so that it's clear for all users.
The approach from #31 of having two distinct "types" is more explicit, and it would be easier to document, but if you think that having it in once is clear enough for you, then I can live with that.

Regarding the checks, can you do something to make the linter happy again? :)

tpluscode:
> The approach from https://github.com/zazuko/barnard59/issues/31 of having two distinct "types" is more explicit, and it would be easier to document, but if you think that having it in once is clear enough for you, then I can live with that.

Yes, and I realise the "double standard" given the similarity in overloading the private key parameter of SFTP step. The difference is that there the overload is actual key vs file path. Here, the literal represents a path in both cases, only possible hidden indirectly inside a variable. Thus, I thought that it could be acceptable

tpluscode:
> (should be mostly OK, since a path looks different enough than a variable name)

Ah, I see now. If you look at the implementation, it does not inspect the value itself. Will only treat the value as variable name when a variable of that name actually exists

PR #145, opened 2023-08-01
fix: missing code to wire-up file loader

tpluscode:
In #142 I forgot to add code which actually sets up the new loader in registry

Here, I did that and added an e2e test which uses the loader


ludovicm67:
Some tests failed for `barnard59-rdf`, can you have a look @tpluscode ?

tpluscode:
Working on it but I don't understand what happened. Seems like unrelated tests start failing 🙄

packages/rdf/lib/localFetch/localFetch.js
```diff
@@ -75,13 +76,17 @@ async function localFetch(
...
+  } catch (e) {
+    const url = isAbsolute(input)
+      ? input
+      : basePath
+        ? pathToFileURL(resolve(basePath, input)).toString()
+        : pathToFileURL(input).toString()
 
-  return fetch(url)
+    return fetch(url)
+  }
```
ludovicm67: I may suggest adding a comment to explain that part, so that we can know why it was done that way in the future.

packages/rdf/lib/localFetch/localFetch.js
```diff
@@ -75,13 +72,14 @@ async function localFetch(
...
-  const url = isAbsolute(input)
-    ? input
-    : basePath
-      ? `file://${resolve(basePath, input)}`
-      : input
+  try {
+    return fetch(new URL(input).toString())
+  } catch {
+    // in case of error, the input must be path string
+    const absolutPath = basePath ? resolve(basePath, input) : input
```
ludovicm67:
```suggestion
    const absolutePath = basePath ? resolve(basePath, input) : input
```

packages/rdf/lib/localFetch/localFetch.js
```diff
@@ -75,13 +72,14 @@ async function localFetch(
...
 
-  return fetch(url)
+    return fetch(pathToFileURL(absolutPath).toString())
```
ludovicm67:
```suggestion
    return fetch(pathToFileURL(absolutePath).toString())
```

ludovicm67 (commented):
Thanks for improving the readability! One very little last thing: I will suggest renaming `absolut` into `absolute`, then I will approve the PR.

ludovicm67 (approved):

PR #157, opened 2023-08-14
feat: filter of CSVW triples

tpluscode:
This appears  in virtually every single pipeline which transforms CSVW and deserves its own module in the CSV package


tpluscode:
RE @ktk @magdasalatka

ktk:
I definitely agree :) On the other hand I still think we could do more with the metadata for tracing, but that deserves a separate discussion. For provenance it would be very useful to at least keep the metadata somewhere for debugging.

PR #170, opened 2023-08-30
Multi root error

giacomociti:
It should work but there's room for improvement. 

A more engineered solution may avoid the dynamic field `skipTrace` and rely on a more specific error class, as suggested for example [here](https://javascript.info/custom-errors) (we should put such a class definition in some core module, where exactly?).

I also would like to improve naming. Whether we keep the dynamic field or we declare some custom error type, the name should describe the kind of error (caused by the user, hence not a system fault) and not how we would like it to be handled (which is not a concern of the code throwing it)

### DIFF
--- a/packages/cli/bin/barnard59.js
+++ b/packages/cli/bin/barnard59.js
@@ -31,10 +31,11 @@ const onError = async err => {
   process.off('SIGTERM', onError)
 
   if (err) {
-    // eslint-disable-next-line no-console
     if (err.skipTrace) {
+      // eslint-disable-next-line no-console
       console.log(err.message)
     } else {
+      // eslint-disable-next-line no-console
       console.log(err)
     }
   }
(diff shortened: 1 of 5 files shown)


packages/cli/bin/barnard59.js
```diff
@@ -31,8 +31,13 @@ const onError = async err => {
   process.off('SIGTERM', onError)
 
   if (err) {
-    // eslint-disable-next-line no-console
-    console.log(err)
+    if (err.skipTrace) {
+      // eslint-disable-next-line no-console
+      console.log(err.message)
+    } else {
```
tpluscode: Yes, I would follow your suggestion and add a derived error class. Let's keep it in this module since it is responsible for this logic.

With that, I think you should move the construction of that message here 

```suggestion
    if (err instanceof MultipleRootsError) {
      const alternatives = err.pipelines.map(x => `\n\t--pipeline ${x}`).join('')
      // eslint-disable-next-line no-console
      console.log(`Multiple root pipelines found. Try one of these:${alternatives}`)
    } else {
```

And in `findPipeline.js` you'd simple throw like `throw new MultipleRootsError({ pipelines: ptr })`

tpluscode (changes requested):

tpluscode (approved):

PR #171, opened 2023-08-31
Migrate validation package

tpluscode:
This is more complex migration than those before because I also update the whole codebase to ESM

That said, I would not put that much emphasis on this library just yet because the near plan is to do a solid refactoring to integrate into `barnard59` CLI


packages/validation/package.json
```diff
@@ -0,0 +1,40 @@
...
+  "devDependencies": {
+    "@jsdevtools/chai-exec": "^2.1.1",
+    "barnard59-base": "^1.2.1",
+    "barnard59-core": "^3.0.1",
+    "barnard59-formats": "^1.4.1",
+    "chai": "^4.3.0",
+    "deep-equal": "^2.0.5",
+    "esmock": "^2.3.8",
+    "sinon": "^15.2.0"
+  },
```
ludovicm67, line 25: I suggest that we also add information about where to find the package, like this:

```suggestion
  },
  "repository": {
    "type": "git",
    "url": "git://github.com/zazuko/barnard59.git",
    "directory": "packages/validation"
  },
  "bugs": {
    "url": "https://github.com/zazuko/barnard59/issues"
  },
  "homepage": "https://github.com/zazuko/barnard59",
```

This is useful to quickly find the repo from npmjs.com and so.

ludovicm67 (approved):
Looks great and tests are passing.

I just made a suggestion about adding extra metadata in the `package.json` file.

PR #172, opened 2023-08-31
fix: import wrong

tpluscode:


tpluscode:
Would have to add tests for this, given how it went unnoticed...

ludovicm67 (dismissed):

packages/ftp/lib/FtpClient.js
```diff
@@ -56,7 +56,7 @@ class FtpClient {
       },
     })
 
-    const finished = require('readable-stream').finished
+    const { finished } = await import('readable-stream')
```
ludovicm67: Can this be imported directly at the top of the file, with the other things from `readable-stream`?

packages/ftp/test/list.test.js
```diff
@@ -1,57 +1,22 @@
...
-      { password: undefined, privateKey: resolve(__dirname, 'support/test.key') },
-    ],
-  ].forEach(([label, serverFactory, additionalOptions]) => {
+  serverConfigurations.forEach(([label, serverFactory, additionalOptions]) => {
     it(`lists files from the given directory ${label}`, async () => {
       await withServer(serverFactory, async server => {
         const options = { ...server.options, ...additionalOptions }
 
         const stream = await list({ pathname: 'data', ...options })
-        const filenames = await getStream.array(stream)
```
ludovicm67, line 54: I needed to change `getStream.array` to `getStreamAsArray` so that it's working ; it was saying that `array` is not a function.
It seems that they did some breaking changes in `get-stream`.

packages/ftp/test/write.test.js
```diff
@@ -0,0 +1,28 @@
...
+    ([label, serverFactory, additionalOptions]) => {
+      it(`writes file at the given path ${label}`, async () => {
+        await withServer(serverFactory, async server => {
+          const options = { ...server.options, ...additionalOptions }
+
+          const stream = await write({ filename: 'tmp/foo.txt', ...options })
+          const input = fs.createReadStream(resolve(__dirname, 'support/data/xyz.txt'))
+          input.pipe(stream)
+
+          await getStream(stream)
```
ludovicm67: @tpluscode I will let you check the right thing to use there, as it's the thing that is in trouble. Maybe `getStreamAsArray`? Or any other method exported by `get-stream`.

packages/ftp/test/write.test.js
```diff
@@ -0,0 +1,28 @@
...
+        })
+
+        strictEqual(fs.readFileSync(resolve(__dirname, 'support/tmp/foo.txt')).toString(), '987\n654')
```
ludovicm67, line 25: For some reason, the `write` is not working ; at least in the tests.
I mean, it's not creating the file.

@tpluscode can I let you check for this?

ludovicm67 (changes requested):
@tpluscode I got changes from `master` branch, so that this one is up-to-date.

I also upgraded dependencies for the `ftp` package, and regenerated the lock file, but now everything is failing.

I think that there was a breaking change somewhere in some of the dependencies that is used everywhere. Can you take a look?

From what I was able to try locally, the strange issues we had for the `ftp` module are gone, but there are some other issues, as you can see in my comments.

packages/ftp/test/list.test.js
tpluscode: Yes, there was that change to `get-stream` along the way

packages/ftp/package.json
```diff
@@ -21,17 +21,18 @@
   "homepage": "https://github.com/zazuko/barnard59",
   "dependencies": {
     "ftp": "^0.3.10",
-    "readable-stream": "^3.6.0",
+    "path-scurry": "^1.10.1",
```
tpluscode: What's that? @ludovicm67

packages/ftp/lib/FtpClient.js
tpluscode: Yea. Actually I don't know why this was loaded dynamically in the first place 🤔

packages/ftp/test/write.test.js
tpluscode: The error I get is because the `stream` is not readable but writable

packages/ftp/package.json
ludovicm67: Without this, I'm getting the following error while running the tests:

```
Error [ERR_MODULE_NOT_FOUND]: Cannot find package 'path-scurry' imported from /***/barnard59/packages/ftp/node_modules/glob/dist/esm/glob.js
    at packageResolve (node:internal/modules/esm/resolve:844:9)
    at moduleResolve (node:internal/modules/esm/resolve:901:20)
    at defaultResolve (node:internal/modules/esm/resolve:1121:11)
    at nextResolve (node:internal/modules/esm/hooks:865:28)
    at resolve (file:///***/barnard59/node_modules/tsm/loader.mjs:1:627)
    at nextResolve (node:internal/modules/esm/hooks:865:28)
    at Hooks.resolve (node:internal/modules/esm/hooks:303:30)
    at handleMessage (node:internal/modules/esm/worker:196:24)
    at Immediate.checkForMessages (node:internal/modules/esm/worker:138:28)
    at process.processImmediate (node:internal/timers:478:21)
```

tpluscode: Strange

ludovicm67: If you remove that change, then the `node_modules` directory and the lock file, and then reinstall everything and if you run `npm run test` for the ftp package, you should be able to reproduce the issue.

Including it was then working well for me. The thing is that I'm not sure if it's only impacting the tests or the whole package, that's why I added it as a dependency and not a dev dependency.

packages/ftp/lib/FtpClient.js
ludovicm67: Ok :) You marked it as resolved ; this mean you will push a change or keep it as it?

tpluscode: Will change

packages/ftp/package.json
tpluscode: Actually, I could not reproduce this

packages/ftp/test/write.test.js
tpluscode: I changed to `finished` but the file is created empty. It looks like the stream created in `FtpClient#write` is not consumed by the `put` function. So far I could not understand why

packages/ftp/package.json
ludovicm67: Even by doing the following from the root:

```sh
rm -rf node_modules package-lock.json
npm i
cd packages/ftp
npm run test
```

?

tpluscode: Yes. Cannot reproduce in node 18 nor in node 20

tpluscode:
@ludovicm67 ok, I managed to resolved the NPM strangeness. Now only the `write` test remains. Have you tried using it in a pipeline with real ftp server?

ludovicm67 (approved):
Changes looks OK, but I don't have time to really try it on a real FTP server.

PR #177, opened 2023-09-15
Simplify step args

giacomociti:


packages/cli/examples/parse-csvw.ttl
```diff
@@ -16,38 +20,17 @@
...
-  ];
-  code:arguments ( <parseMetadata> ).
-
-<serialize> a p:Step;
-  code:implementedBy [ a code:EcmaScriptModule;
-    code:link <node:barnard59-formats/ntriples.js#serialize>
+    p:stepList ( 
+      [ fs:createReadStream ("filename"^^p:VariableName) ] 
+      [ csvw:parse (<parseMetadata>) ]
+      [ ntriples:serialize () ]
```
tpluscode, line 26: I really like how it looks with the prefixes 🤩

packages/cli/lib/pipeline.js
```diff
@@ -2,15 +2,71 @@ import { dirname, resolve } from 'path'
...
+        ops.set(x.term, link)
+      })
+  }
+
+  return ops
+}
+
+export const desugarWith = context => dataset => {
+  const { knownOperations } = context
+  const ptr = clownface({ dataset })
```
tpluscode: I recommend using clownface via the environment

```suggestion
  const ptr = rdf.clownface({ dataset })
```

packages/cli/lib/pipeline.js
```diff
@@ -2,15 +2,71 @@ import { dirname, resolve } from 'path'
...
+  let n = 0
+  ptr.has(ns.rdf.type, ns.p.Pipeline)
+    .out(ns.p.steps)
+    .out(ns.p.stepList)
+    .forEach(x => {
```
tpluscode: Wonder if it it needs to be that precise. Might go straight for `stepList`
```suggestion
  ptr.has(ns.p.stepList)
    .out(ns.p.stepList)
    .forEach(x => {
```

packages/cli/lib/pipeline.js
```diff
@@ -2,15 +2,71 @@ import { dirname, resolve } from 'path'
...
+      for (const step of x.list()) {
+        if (step.out(ns.code.implementedBy).terms.length === 0) {
+          // when there is no implementation, we expect a known operation
+          const [quad] = step.dataset.match(step.term)
+          const knownStep = knownOperations.get(quad.predicate)
```
tpluscode: I think there is the slight possibility that `quad` would be undefined. Might guard against that

```suggestion
          const [quad] = step.dataset.match(step.term)
          const knownStep = knownOperations.get(quad?.predicate)
```

packages/cli/lib/pipeline.js
```diff
@@ -2,15 +2,71 @@ import { dirname, resolve } from 'path'
...
+export const desugarWith = context => dataset => {
+  const { knownOperations } = context
+  const ptr = clownface({ dataset })
+  let n = 0
+  ptr.has(ns.rdf.type, ns.p.Pipeline)
+    .out(ns.p.steps)
+    .out(ns.p.stepList)
+    .forEach(x => {
+      for (const step of x.list()) {
+        if (step.out(ns.code.implementedBy).terms.length === 0) {
```
tpluscode: FYI, for that purpose I published the package [is-graph-pointer](https://npm.im/is-graph-pointer)

```
import { isGraphPointer } from 'is-graph-pointer'
```

```suggestion
        if (!isGraphPointer(step.out(ns.code.implementedBy))) {
```

packages/cli/lib/pipeline.js
```diff
@@ -2,15 +2,71 @@ import { dirname, resolve } from 'path'
...
+              step.addOut(ns.code.arguments, args)
+            }
+            step.addOut(ns.rdf.type, ns.p.Step)
+            const moduleNode = ptr.blankNode(`impl_${n++}`)
+            moduleNode.addOut(ns.rdf.type, ns.code.EcmaScriptModule)
+            moduleNode.addOut(ns.code.link, knownStep)
+            step.addOut(ns.code.implementedBy, moduleNode)
+          }
+        }
+      }
```
tpluscode: I think I would also handle the other branch of this conditional. If you'd add the logger to the context and at least log in debug level but maybe even warn actually, as it's important that step would not be found. Also, I personally prefer less nesting

```suggestion
          if (!knownStep) {
            log.debug(`...`)
          }

          const args = step.out(quad.predicate)
          step.deleteOut(quad.predicate)
          // keep args only if non-empty
          if (!ns.rdf.nil.equals(args.term)) {
            step.addOut(ns.code.arguments, args)
          }
          step.addOut(ns.rdf.type, ns.p.Step)
          const moduleNode = ptr.blankNode(`impl_${n++}`)
          moduleNode.addOut(ns.rdf.type, ns.code.EcmaScriptModule)
          moduleNode.addOut(ns.code.link, knownStep)
          step.addOut(ns.code.implementedBy, moduleNode)  
        }
      }
```

tpluscode (changes requested):

packages/cli/lib/pipeline.js
```diff
@@ -12,10 +12,11 @@ const discoverOperations = async () => {
...
       .has(rdf.ns.rdf.type, ns.p.Operation)
-      .forEach(x => { // distinguish EcmaScript and EcmaScriptModule?
-        const link = x.out(ns.code.implementedBy).out(ns.code.link).term
-
-        ops.set(x.term, link)
+      .forEach(x => {
+        const impl = x.out(ns.code.implementedBy)
+        const type = impl.out(ns.rdf.type).term
+        const link = impl.out(ns.code.link).term
+        ops.set(x.term, { type, link })
```
tpluscode: Maybe `x` is not greatest variable name :D
```suggestion
      .forEach(operation => {
        const impl = operation.out(ns.code.implementedBy)
        const type = impl.out(ns.rdf.type).term
        const link = impl.out(ns.code.link).term
        ops.set(operation.term, { type, link })
```

packages/cli/lib/pipeline.js
giacomociti: the step list may include a pipeline. In this case the operation is not found but we should not have a warning.
Not sure if it's worth having an exhaustive conditional logic

tpluscode: I'd say it's worth it. Silently moving on will cause "strange" error. The reporting of steps which fail to load is not that great right now

With what you said in mind, you need to check that it's not in fact a pipeline and log then

giacomociti: maybe we can look for operations only when the list element is neither an instance of Step nor an instance of Pipeline.
Then, we can raise a warning if the operation is not found

tpluscode (approved):

test/e2e/definitions/foreach/csv-duplicate.ttl
giacomociti, line 1: mixing simplified syntax and regular steps in the same list is not that ugly after all (as long as the step definition is not inlined)

tpluscode: Yes, probably something we can live with

PR #183, opened 2023-09-27
Create s3 package

ludovicm67:
This PR creates a package that includes the following methods to manipulate S3 objects:
- `getObject`
- `putObject`

I tried to use those functions locally against a S3 endpoint, and it is working as expected.
In the future I will try to add relevant tests.

I already included JSDoc annotations to have some useful autocompletion and type checking.


ludovicm67:
I added a README file and tests for the `getObject` function.

ludovicm67:
I added tests in dedicated files.

packages/s3/cmd/getObject.js
```diff
@@ -0,0 +1,45 @@
...
+  const client = newClient(generateConfig({ accessKeyId, secretAccessKey, ...s3Config }))
+  const input = {
+    Bucket: bucket,
+    Key: key,
+  }
+  const command = new GetObjectCommand(input)
+  const data = await client.send(command)
+
+  if (!data || !data.Body) {
+    throw new Error('No body')
```
tpluscode: Let's make this message more informative

.changeset/olive-phones-drop.md
```diff
@@ -0,0 +1,5 @@
+---
+"barnard59-s3": patch
```
tpluscode:
```suggestion
"barnard59-s3": minor
```

maybe

packages/s3/lib/client.js
```diff
@@ -0,0 +1,41 @@
...
+  if (accessKeyId && secretAccessKey) {
+    config.credentials = {
+      accessKeyId,
+      secretAccessKey,
+    }
+  }
+
+  // set a default region if none is provided
+  if (!s3Config.region) {
+    config.region = 'us-east-1'
```
tpluscode: Why us-east-1 in particular?

tpluscode (changes requested):
One last thing: please add a row in the root README.md

packages/s3/lib/client.js
ludovicm67: It's the default region that is used in general if no one is specified.
But for most cases, the region is not required.

Maybe I can use the `process.env.AWS_DEFAULT_REGION || 'us-east-1'` value, so that it can support AWS directly out of the box (some tools supports using the [environment variables from AWS](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-envvars.html)).

tpluscode: Not a bad idea if that's natural to S3 users

tpluscode (approved):

PR #189, opened 2023-10-13
feat: glob warn when not matched

tpluscode:


packages/base/glob.js
```diff
@@ -6,6 +6,7 @@ import { Readable } from 'readable-stream'
 import tracer from './lib/tracer.js'
 
 function glob({ pattern, ...options }) {
+  const { logger } = this
```
giacomociti, line 9: is it worth covering also the cases in which `logger` (or even `this`) is undefined?

giacomociti (approved):

packages/base/glob.js
tpluscode: No. That should never happen

ludovicm67 (approved):

Issue #161, opened 2023-08-22
Some errors are emitted to stdout instead of stderr


tpluscode:

giacomociti:
browsing history I noticed commit 9bb6520 changed `console.error(err)` to `console.log(err)`. Maybe we should just revert to using `console.error`?

tpluscode:
Good catch. There should be no `console` at all

giacomociti:
not sure about removing console altogether, at least the error message for multiple roots is useful (see the related PR)

ktk:
Not sure if that changed but console.log was or is synchronous so it's better to use an abstraction for logging that *can* write to console but doesn't do so unless we tell it to. Not sure what's the latest and greates logger though.

giacomociti:
yes, we have the proper logging abstractions in place and all the other telemetry stuff. My guess is that this is a special case because we are at the boundaries of the system, where even the logging abstractions may not be available yet.

PR #200, opened 2023-11-09
fix: restore log to stderr

giacomociti:


giacomociti:
should address #161

tpluscode (approved):
Oh, so the only ones left are in the CLI. Then they are not the problem of stream errors being written to the output. Anyway, let's merge this

PR #201, opened 2023-11-10
Current manifest

giacomociti:
The current project is included in manifest discovery, enabling simplified syntax also for operations defined in the manifest of the current project (as exemplified in the e2e tests). Maybe this is even more useful to enable package-specific CLI commands defined in the current project's manifest.

### DIFF
--- a/packages/cli/lib/discoverManifests.js
+++ b/packages/cli/lib/discoverManifests.js
@@ -1,4 +1,5 @@
 import * as module from 'module'
+import path from 'path'
 import fs from 'fs'
 import findPlugins from 'find-plugins'
 import rdf from 'barnard59-env'
@@ -13,6 +14,15 @@ export default async function * () {
       return packagePattern.test(pkg.name) && hasManifest(pkg.name)
     },
   })
+
+  const dir = process.cwd()
+  if (hasManifest(dir)) {
+    yield {
+      name: path.basename(dir),
+      manifest: rdf.clownface({ dataset: await rdf.dataset().import(rdf.fromFile(`${dir}/manifest.ttl`)) }),
+      version: require(`${dir}/package.json`).version,
+    }
+  }
 
   for (const { pkg } of packages) {
     const { version } = require(`${pkg.name}/package.json`)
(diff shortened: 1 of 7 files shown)


packages/cli/lib/discoverManifests.js
```diff
@@ -14,6 +15,15 @@ export default async function * () {
     },
   })
 
+  const dir = process.cwd()
```
tpluscode: I would not rely on `cwd`. Maybe use [`pkg-dir`](https://github.com/sindresorhus/pkg-dir) to find the package root

```suggestion
  const dir = await packageDirectory()
```

tpluscode (changes requested):
I take back what I said previously. It is going to be super useful to move local operations to a manifest and have super clean pipeline definitions 💪 

Then let's add the `code:imports` loader so that you'd be able to split a manifest in multiple files 😎

packages/cli/lib/discoverManifests.js
```diff
@@ -14,6 +15,15 @@ export default async function * () {
...
+  if (hasManifest(dir)) {
+    yield {
+      name: path.basename(dir),
+      manifest: rdf.clownface({ dataset: await rdf.dataset().import(rdf.fromFile(`${dir}/manifest.ttl`)) }),
+      version: require(`${dir}/package.json`).version,
+    }
+  }
```
tpluscode: Ah, might also use name from package.json itself
```suggestion
  if (hasManifest(dir)) {
    const { name, version } = require(`${dir}/package.json`)
    yield {
      name,
      manifest: rdf.clownface({ dataset: await rdf.dataset().import(rdf.fromFile(`${dir}/manifest.ttl`)) }),
      version,
    }
  }
```

packages/cli/lib/discoverManifests.js
giacomociti: ah, ok. I took the usage of `cwd` from the implementation of `findPlugins`

tpluscode (approved):

PR #202, opened 2023-11-13
cube package

giacomociti:
A new _cube_ package is created, including operations previously available in the _rdf_ package as well as commands (pipelines) to validate observations and their constraints.


packages/sparql/manifest.ttl
```diff
@@ -7,12 +7,12 @@
   rdfs:label "SPARQL Select";
   rdfs:comment "Runs the given CONSTRUCT query against the given endpoint parses the result.";
   code:implementedBy [ a code:EcmaScriptModule;
-    code:link <node:barnard59-sparql#construct.js#default>
+    code:link <node:barnard59-sparql/construct.js#default>
```
tpluscode, line 10: That was a mistake before?

packages/cube/package.json
```diff
@@ -0,0 +1,49 @@
+{
+  "name": "barnard59-cube",
+  "version": "2.0.0",
```
tpluscode:
```suggestion
  "version": "0.0.0",
```

packages/cube/package.json
```diff
@@ -0,0 +1,49 @@
...
+  "description": "RDF cubes for Linked Data pipelines",
```
tpluscode: Maybe

```suggestion
  "description": "Build and check RDF cubes in Linked Data pipelines",
```

packages/cube/package.json
```diff
@@ -0,0 +1,49 @@
...
+    "barnard59-base": "^2.0.0",
+    "barnard59-formats": "^2.0.0",
+    "barnard59-http": "^2.0.0",
+    "barnard59-rdf": "^2.0.0",
+    "barnard59-sparql": "^2.0.0",
+    "barnard59-validate-shacl": "^0.3.8"
+  },
+  "engines": {
+    "node": ">= 14.0.0"
+  },
```
tpluscode:
```suggestion
```

I think Ludovic removes most of those

packages/cube/package.json
```diff
@@ -0,0 +1,49 @@
...
+  "dependencies": {
+    "@zazuko/env-node": "^1.0.0",
+    "external-merge-sort": "^0.1.2",
+    "lodash": "^4.17.21",
+    "rdf-dataset-ext": "^1.0.1",
```
tpluscode: Please use `this.env` provided by step factories (re #195). Check the snippets in #198

```suggestion
    "external-merge-sort": "^0.1.2",
    "lodash": "^4.17.21",
```

packages/cube/package.json
```diff
@@ -0,0 +1,49 @@
...
+    "rdf-literal": "^1.3.0",
+    "rdf-stream-to-dataset-stream": "^1.0.0",
+    "readable-stream": "^3.6.0"
```
tpluscode:
```suggestion
    "readable-stream": "3 - 4"
```

This can help avoid duplicate resolutions

packages/cube/manifest.ttl
```diff
@@ -0,0 +1,54 @@
+@base <https://barnard59.zazuko.com/> .
+@prefix b59: <https://barnard59.zazuko.com/vocab#> .
+@prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#> .
+@prefix code: <https://code.described.at/> .
+@prefix p: <https://pipeline.described.at/> .
+
+# renamed from <https://barnard59.zazuko.com/operations/rdf/cube.js#buildCubeShape>
```
tpluscode, line 7: Maybe less discoverable here, I would add an exact example of how consumers should update their pipelines in the changeset. Like a diff snippet

packages/cli/lib/discoverManifests.js
```diff
@@ -10,6 +10,7 @@ const require = module.createRequire(import.meta.url)
 export default async function * () {
   const packages = findPlugins({
     includeDev: true,
+    includePeer: true,
```
tpluscode, line 13: Should add a changeset to update the CLI version

packages/cube/README.md
```diff
@@ -0,0 +1,90 @@
...
+a [concise bounded description](https://docs.stardog.com/query-stardog/#describe-queries) of the `cube:Constraint` part of a given cube.
+
+```bash
+npx barnard59 run ./pipeline/cube-validation.ttl \
+    --pipeline http://barnard59.zazuko.com/pipeline/cube-validation/fetch-cube-constraint \
+    --variable cube=https://agriculture.ld.admin.ch/agroscope/PRIFm8t15/2 \
+    --variable endpoint=https://int.lindas.admin.ch/query 
+```
+
+Taking advantage of [package-specific commands](https://data-centric.zazuko.com/docs/workflows/reference/cli/#package-specific-commands), we can express the same as:
```
tpluscode: Is it even worth showing the `barnard59 run` usage in the readme?

packages/cube/README.md
```diff
@@ -0,0 +1,90 @@
...
+
+The pipeline reads the observations from `stdin`, allowing input from a local file (as in the following example) as well as from the output of the `fetch-cube-observations` pipeline.
+
+```bash
+cat myObservations.ttl \
+| npx barnard59 cube check-observations \
+    --constraint myConstraint.ttl
+```
+
+To enable validation, the pipeline adds to the constraint a `sh:targetClass` property with value `cube:Observation` (assuming that each observation has an explicit property `rdf:type` with value `cube:Observation`).
```
tpluscode:
```suggestion
To enable validation, the pipeline adds to the constraint a `sh:targetClass` property with value `cube:Observation`, requiring that each observation has an explicit `rdf:type`.
```

packages/cube/lib/batch.js
```diff
@@ -0,0 +1,30 @@
...
+  for await (const array of chunkObjectsBySize(size, iterable)) {
+    const batch = rdf.dataset()
+    for (const dataset of array) {
+      batch.addAll(dataset)
+    }
+    yield batch
+  }
+}
+
+export const batch = size => Duplex.from(iterable => chunkBySize(Number(size), iterable))
```
tpluscode: Is `chunkBySize` ever imported anywhere? I ask because you'd want to pass RDF env from the step factory

```suggestion
export async function * chunkBySize({ env }, size, iterable) {
  for await (const array of chunkObjectsBySize(size, iterable)) {
    const batch = env.dataset()
    for (const dataset of array) {
      batch.addAll(dataset)
    }
    yield batch
  }
}

export const batch = function (size) => {
  return Duplex.from(iterable => chunkBySize(this, Number(size), iterable))
}
```

This is an example of how you'd use the env from barnard context

packages/cube/lib/namespaces.js
```diff
@@ -0,0 +1,6 @@
+import $rdf from '@zazuko/env-node'
+
+const cube = $rdf.namespace('https://cube.link/')
+const { rdf, rdfs, sh, xsd, _void, dcat, schema, dcterms } = $rdf.ns
+
+export { cube, rdf, rdfs, sh, xsd, _void, dcat, schema, dcterms }
```
tpluscode: You can remove this once you use `this.env.ns` which contains all namespaces, including `cube`

packages/cube/lib/sort.js
```diff
@@ -0,0 +1,15 @@
+import { Readable, Duplex } from 'stream'
+import { sort, compareOn, createStore } from 'external-merge-sort'
+import rdf from '@zazuko/env-node'
+
+async function write(chunk, filename) {
+  await rdf.toFile(Readable.from(chunk), filename)
+  return rdf.fromFile(filename)
+}
```
tpluscode: Not exported. Unused?

packages/cube/pipeline/cube-validation.ttl
```diff
@@ -0,0 +1,229 @@
...
+@prefix base: <https://barnard59.zazuko.com/operations/base/> .
+@prefix n3: <https://barnard59.zazuko.com/operations/formats/n3/> .
+@prefix ntriples: <https://barnard59.zazuko.com/operations/formats/ntriples/> .
+
+@base <http://barnard59.zazuko.com/pipeline/cube-validation/> .
+
+_:endpoint a p:Variable ;
+  p:name "endpoint" ;
+  rdfs:label "SPARQL endpoint" ;
+  p:value "https://lindas.admin.ch/query"
```
tpluscode: Not sure there should be a default value

packages/validate-shacl/manifest.ttl
```diff
@@ -9,3 +9,10 @@
           code:implementedBy [ a code:EcmaScriptModule;
                               code:link <node:barnard59-validate-shacl/validate.js#shacl>
                              ].
+
+<report> a p:Operation, p:Readable;
+          rdfs:label "validate rdf";
+          rdfs:comment "Validates a RDF graph against a set of conditions specified in a SHACL document";
```
tpluscode: This should be different from the description of `validate`. Else, it's not clear what is the difference

packages/cube/manifest.ttl
```diff
@@ -0,0 +1,54 @@
...
+  rdfs:label "Retrieves cube observations from SPARQL endpoint" ;
+  b59:source "barnard59-cube/pipeline/cube-validation.ttl" ;
+  b59:pipeline <http://barnard59.zazuko.com/pipeline/cube-validation/fetch-cube-observations> 
+.
+
+<command/cube/check-observations>
+  a b59:CliCommand ;
+  b59:command "check-observations" ;
+  rdfs:label "Validate input observations against cube constraint" ;
+  b59:source "barnard59-cube/pipeline/cube-validation.ttl" ;
```
tpluscode: I'd place each pipeline in its own file. Are they all in `cube-validation.ttl` to avoid duplication?

packages/cube/pipeline/cube-validation.ttl
```diff
@@ -0,0 +1,229 @@
...
+
+<fetch-cube-constraint> a p:Pipeline ;
+  p:variables [ p:variable _:endpoint, _:cube ] ;
+  p:steps
+    [
+      p:stepList
+        (
+          _:queryConstraint
+          [ ntriples:serialize () ]
+          _:stdout
```
tpluscode: You don't need an explicit `stdout` step. That is the default behavior

```suggestion
```

packages/cube/pipeline/cube-validation.ttl
```diff
@@ -0,0 +1,229 @@
...
+
+
+<fetch-cube-constraint> a p:Pipeline ;
+  p:variables [ p:variable _:endpoint, _:cube ] ;
+  p:steps
+    [
+      p:stepList
+        (
+          _:queryConstraint
+          [ ntriples:serialize () ]
```
tpluscode: Actually, is it even necessary to parse and then serialize again? Looks to me that you'd be simply able to stream the query results if you request n-triples

packages/cube/pipeline/cube-validation.ttl
```diff
@@ -0,0 +1,229 @@
...
+          [code:name "maxErrors" ; code:value "maxViolations"^^p:VariableName ] 
+        ]
+        [ base:flatten () ]
+        [ ntriples:serialize () ]
+        _:stdout
+      )
+    ]
+.
+
+# relying on cbd is not ideal (vendor specific)
```
tpluscode: :+1:
We'll replace this with `CONSTRUCT` later

packages/cube/pipeline/cube-validation.ttl
```diff
@@ -0,0 +1,229 @@
...
+. 
+
+_:batch a p:Step ;
+  code:implementedBy 
+    [ 
+      a code:EcmaScriptModule ;
+      code:link <file:../lib/batch.js#batch> ;
+    ] ;
+  code:arguments ("batchSize"^^p:VariableName)
+.
```
tpluscode: Isn't there some step like this in `barnard59-base`?

packages/cube/pipeline/cube-validation.ttl
```diff
@@ -0,0 +1,229 @@
...
+
+_:readConstraint a p:Step ;
+  code:implementedBy 
+    [ 
+      a code:EcmaScriptModule ;
+      code:link <node:rdf-utils-fs#fromFile> ;
+    ] ;
+  code:arguments ("constraint"^^p:VariableName)
+.
```
tpluscode: I added a suggestion to remove `rdf-utils-fs` because `@zazuko/env-node` provides that functionality. 

This should work

```suggestion
_:readConstraint a p:Step ;
  code:implementedBy """
  function (path) {
    return this.env.fromFile(path)
  }
  """^^code:EcmaScript ;
  code:arguments ("constraint"^^p:VariableName)
.
```

tpluscode (changes requested):

packages/sparql/manifest.ttl
giacomociti: guess so

packages/cube/package.json
giacomociti: apparently I'm still confused by versioning  :)

packages/cube/package.json
giacomociti: I would be more confident if you could apply these changes

packages/cube/README.md
giacomociti: ok, will remove it

packages/cube/lib/batch.js
giacomociti: not used elsewhere, we can remove the `export` and use the env from barnard context

packages/sparql/manifest.ttl
tpluscode: Then we'd also need to bump the  version here

packages/cube/lib/sort.js
giacomociti: used a few lines below:
`const store = createStore(write, '.nt')`

This is the function that spills to temporary files on disk when the number of items to sort exceeds a threshold

packages/cube/package.json
tpluscode: This is the current, or initial, version. It could be anything, for sure. And then you added the changeset which bumps major. So with `2.0`, we'd publish `v3` to NPM when this is merged

packages/cube/pipeline/cube-validation.ttl
giacomociti: we can remove it

packages/cube/lib/sort.js
tpluscode: Ah, ok. Code coloring misled me

packages/cube/manifest.ttl
giacomociti: yes but the actual duplication would be very little

packages/cube/pipeline/cube-validation.ttl
giacomociti: not sure about this, I'll try again but I think I already tried and got no output

packages/cube/pipeline/cube-validation.ttl
giacomociti: that would mean to use a generic http fetch instead of the barnard59-sparql operations? Because my understanding is that the 'construct' operation returns a stream of quads

packages/cube/pipeline/cube-validation.ttl
giacomociti: apparently not, in fact I would consider adding one (or even two: one for datasets and another for generic objects)

packages/cube/pipeline/cube-validation.ttl
giacomociti: I will try

packages/sparql/manifest.ttl
giacomociti: done

packages/cli/lib/discoverManifests.js
giacomociti: done

packages/cube/README.md
giacomociti: done

packages/cube/pipeline/cube-validation.ttl
giacomociti: done

packages/validate-shacl/manifest.ttl
giacomociti: done

packages/cube/pipeline/cube-validation.ttl
tpluscode: https://github.com/zazuko/barnard59/issues/206

tpluscode (approved):
I think we're good here 👍

PR #207, opened 2023-11-20
refactor: no stdout

tpluscode:


giacomociti:
I think there are two independent problems (and both are not covered by tests because it's not easy):
 - writing to stdout (only `check-observation` is actually doing it)
 - cleanup of temp files produced by `sortBySubject`

The latter is due to the fact that, when an error is thrown in a step, the previous ones don't get the chance to do any finalization.
Apparently, using the `stream.pipeline` function instead of the `pipe` method would provide the possibility, but this would be a risky change to the core of barnard59. The only reasonable workaround for me was to avoid [breaking](https://github.com/zazuko/barnard59/blob/c7e328a04013104d9178fb29773b319ccba236f4/packages/validate-shacl/report.js#L10) the stream (using `continue` instead of `break` and throw the validation error only once all input data has flown through the pipeline.

giacomociti:
let me add that the (lack of) finalization problem is not strictly due to the error being thrown but is more general: when a step ends, the steps before it apparently don't get any notification.

packages/cube/test/validation.pipeline.test.js
```diff
@@ -27,7 +27,29 @@ describe('cube validation pipeline', function () {
     const result = shell.exec(command, { silent: true, cwd })
 
     strictEqual(result.code, 1)
-    expect(result.stderr).to.match(/^Error: At least 1 violations found/)
+    expect(result.stderr).to.match(/1 violations found/)
+    ok(result.stdout.includes('_:report <http://www.w3.org/ns/shacl#conforms> "false"^^<http://www.w3.org/2001/XMLSchema#boolean>'))
+  })
+
+  it('should run check-cube-observations pipeline without error when maxViolations is not exceeded', () => {
```
giacomociti: I'm not convinced this behaviour is useful: the main purpose of maxViolation is to avoid bloating the output (usually a few examples are enough to understand what's wrong with the input) but the cli should succeed only if there are no violations at all, allowing to break a script in which the validation pipeline is used to ensure that all the input is conformant

tpluscode: You're right. Will change that. Then maybe we need a much higher default. Like 100 or more?

tpluscode:
As a final touch, I added minimal code required to handle `this.error` when called multiple times and from sub-pipelines. The lazy error would no longer be set directly but passed with a function

PR #209, opened 2023-11-24
Rename SHACL package

tpluscode:
I also  headers in changelog files


README.md
giacomociti, line 1: should we add the cube package here already?

giacomociti (approved):

README.md
tpluscode: Good point. I added it and also updated the readme in shacl package

PR #216, opened 2023-11-27
Add support for streams for `barnard59-s3`

ludovicm67:
This PR adds support for stream for `barnard59-s3`.

This allows users to create a stream from a S3 object, and put a stream into a S3 object.


ludovicm67:
I'm removing the `putStreamObject` function, as it cannot be working as we might expect.

Or if someone has a way to do it, feel free to take over that PR and add support to it.

Until then, the recommended way is to write into a file on the current file system first, and use the `putObject` function to send that file into the S3 bucket.

packages/s3/.gitignore
```diff
@@ -1 +1,2 @@
 test/results
+coverage/
```
tpluscode, line 2: Already ignored in top `.gitignore`. Is that not covering it?

packages/s3/test/lib.test.js
```diff
@@ -0,0 +1,101 @@
...
+      })
+
+      try {
+        // Pass the mock stream to your function
+        await toString(mockStream)
+        throw new Error('Error was not thrown as expected')
+      } catch (error) {
+        // Assert the error
+        strictEqual(error.message, 'Test error')
+      }
```
tpluscode: I highly recommend using chai with plugins. For example, using `chai-as-promised`, this could become

```suggestion
      await expect(toString(mockStream)).to.be.eventually.rejectedWith('Test error')
```

packages/s3/manifest.ttl
```diff
@@ -16,3 +16,10 @@
   code:implementedBy [ a code:EcmaScriptModule;
     code:link <node:barnard59-s3/cmd/putObject.js#default>
   ].
+
+<getStreamObject> a p:Operation;
+  rdfs:label "Get S3 object as stream";
+  rdfs:comment "Get an object from a S3 bucket as a Readable stream.";
+  code:implementedBy [ a code:EcmaScriptModule;
+    code:link <node:barnard59-s3/cmd/getStreamObject.js#default>
+  ].
```
tpluscode: What do you think about calling this `s3:getObject.stream`?

packages/s3/manifest.ttl
```diff
@@ -16,3 +16,10 @@
   code:implementedBy [ a code:EcmaScriptModule;
     code:link <node:barnard59-s3/cmd/putObject.js#default>
   ].
+
+<getStreamObject> a p:Operation;
```
tpluscode: I see this missing above but I think it's necessary

```suggestion
<getStreamObject> a p:Operation, p:Readable;
```

packages/s3/lib/streams.js
```diff
@@ -34,3 +34,63 @@ export const createWritableStream = (path) => {
...
+
+    stream.on('end', () => {
+      resolve(data.join(''))
+    })
+
+    stream.on('error', (error) => {
+      reject(error)
+    })
+  })
+}
```
tpluscode: Why not use `get-stream` package?

packages/s3/lib/streams.js
```diff
@@ -34,3 +34,63 @@ export const createWritableStream = (path) => {
...
+      // Handle any errors
+      nodeStream.emit('error', err)
+    })
+  }
+
+  // Start pumping data
+  pump()
+
+  return nodeStream
+}
```
tpluscode: Would is be possible to use `Readable.fromWeb`?
Similarly, could `toWritableStream` be replaced with `Writable.toWeb`?

.changeset/fifty-guests-kick.md
```diff
@@ -0,0 +1,11 @@
+---
+"barnard59-s3": minor
+---
+
+Allow `getObject` and `putObject` operations to be used inside a barnard59 pipeline, as they are now returning a stream.
+They can be used at any place in the list of steps, as it will not change anything in the stream.
```
tpluscode, line 6: I'm a little uneasy about this. Are those just no-op steps with side effects? I don't think we have any other step that behave like this

Can you show me an example usage in a pipeline?

packages/s3/cmd/getStreamObject.js
```diff
@@ -0,0 +1,40 @@
...
+    Key: key,
+  }
+  const command = new GetObjectCommand(input)
+  const data = await client.send(command)
+
+  if (!data || !data.Body) {
+    throw new Error('There was an issue while fetching the requested S3 object.')
+  }
+
+  const stream = data.Body.transformToWebStream()
```
tpluscode: What is the type of `data.Body`?

tpluscode (changes requested):

packages/s3/lib/streams.js
ludovicm67: As it's only used for tests and that it's a simple function, I don't think it's required to add extra packages ; but I can add it.

tpluscode: It is already wildly used in other packages also for tests

packages/s3/manifest.ttl
ludovicm67: With the dot in the middle, it looks a bit strange?
But I can call it `getObjectStream` if you think it's better?

packages/s3/.gitignore
ludovicm67: I tend to be explicit in each package, so that if we edit the one at the root, we are sure to not send this directory by mistake.

I can remove it.

packages/s3/manifest.ttl
tpluscode: Trying find a pattern, some step use slash. Like `<json/parse>` and `<json/stringify>`

The upside of dot is that it does not need to be escaped in prefixed form

packages/s3/test/lib.test.js
ludovicm67: This would add two dev dependencies: `chai` and `chai-as-promised`. It does not add much value there, but I can add them.

packages/s3/.gitignore
tpluscode: No problem, let's keep it.
No other package needed that anyway, because we're only collecting coverage from workspace root and there are no explicit `coverage` scripts anywhere

packages/s3/test/lib.test.js
tpluscode: Up to you. `chai` is already used. The plugin isn't, yet but overall I find it much cleaner that fighting with the limited capabilities of `assert` and plain mocha

.changeset/fifty-guests-kick.md
ludovicm67: You can start a pipeline with a `getObject`, then just use that file for other steps, and it will be working.

Similarly, if you want to store data in S3, you have the n-1 th step that writes the data into a file, and you can use `putObject` to store that file into a S3 bucket.

By side effects, the only one is that it's writing a file on the file system (`getObject`) or in a S3 bucket (`putObject`). No other side effect ; if so, it's a bug that should be fixed.

Without using streams for them was, it was impossible to add those steps in any pipeline, and then we can just remove the s3 package.

packages/s3/cmd/getStreamObject.js
ludovicm67: This is the type of `data.Body`: https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/Package/-smithy-types/TypeAlias/StreamingBlobPayloadOutputTypes/

packages/s3/lib/streams.js
ludovicm67: Done!

packages/s3/manifest.ttl
ludovicm67: If `getObjectStream` is not OK, I suggest going with `getObject/stream`, so that it has the same pattern as the other operations.
Importing it as a dedicated prefix is not an issue.

I just did a commit to go in that direction.

But now the pattern of that operation name doesn't reflect the name of the file…

packages/s3/lib/streams.js
ludovicm67: > Would is be possible to use `Readable.fromWeb`?

It doesn't seem to work.
If you find another solution, we can try, else I suggest keeping this for now.

> Similarly, could `toWritableStream` be replaced with `Writable.toWeb`?

For this one, I was able to replace it.

packages/s3/test/lib.test.js
ludovicm67: The pattern is not too complex and since the test files are small enough and still readable, we can keep it that way for now, I would say.

ludovicm67:
@tpluscode I finally went through the different points :)
Can you check if changes are still required?

packages/s3/lib/streams.js
tpluscode: > It doesn't seem to work.

Do you remember what failed? I simply replaced `toReadable` with `Readable.fromWeb` and tests passed

packages/s3/manifest.ttl
tpluscode: > But now the pattern of that operation name doesn't reflect the name of the file…

I think we can accept that :)

packages/s3/lib/streams.js
ludovicm67: I had a test failing (the one for the `toReadable` function ; if I replaced the call with `Readable.fromWeb` it was giving the following error: `TypeError [ERR_INVALID_ARG_TYPE]: The "readableStream" argument must be an instance of ReadableStream. Received an instance of Object`).
So I just removed the test ; the rest is passing.

I also needed to add a `// @ts-ignore` comment so that it doesn't complain anymore with types issues.

tpluscode (approved):

PR #221, opened 2023-11-28
Hotfix oom

giacomociti:
both changes are needed to avoid out of memory errors. 
The error in manifest would lead to a single, huge chunk. 
The reuse of the validator instance maybe is worth further investigation to understand why it apparently leaks memory.
But creating a new validator instance for each chunk is sufficient here

### DIFF
--- a/.changeset/fresh-deers-switch.md
+++ b/.changeset/fresh-deers-switch.md
@@ -2,4 +2,4 @@
 "barnard59-shacl": patch
 ---
 
-reduce memory pressure
+prevent potential memory leaks by creating a fresh validator for each chunk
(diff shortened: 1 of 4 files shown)


.changeset/fresh-deers-switch.md
```diff
@@ -0,0 +1,5 @@
+---
+"barnard59-shacl": patch
+---
+
+reduce memory pressure
```
tpluscode: I'd be more specific to what was fixed and how

.changeset/stale-feet-report.md
```diff
@@ -0,0 +1,5 @@
+---
+"barnard59-rdf": patch
+---
+
+fix manifest entry for splitDataset (bySubject and byType)
```
tpluscode, line 5: oops 🙊

tpluscode (approved):

PR #226, opened 2023-11-29
Constraint builders

giacomociti:
Refactoring of shape creation to refine it and to actually reflect the data.
Another aim is to make this usable also from cube creator (which currently has an almost identical copy of the shape creation classes).

- added limit on `sh:in` size (as already available on cube creator)
- moved range constraints (`sh:minInclusive` and `sh:maxInclusive`) inside `sh:or` in case of multiple data types
- skip range constraints in case of parsing issues
- values of `sh:nodeKind` corresponding to the actual data (see https://github.com/zazuko/cube-creator/issues/1469 and https://github.com/zazuko/cube-link/issues/109)


packages/cube/lib/cube/buildCubeShape/Constraints.js
```diff
@@ -0,0 +1,210 @@
...
+    this.builders.forEach(builder => builder.add(object))
+  }
+
+  build(ptr) {
+    this.builders.forEach(builder => builder.build(ptr))
+  }
+}
+
+export class NodeKindConstraintBuilder {
+  #termTypes = new Set()
```
tpluscode: `#` prefix is a convention for private fields?

packages/cube/lib/cube/buildCubeShape/Constraints.js
```diff
@@ -0,0 +1,210 @@
...
+  }
+
+  add(object) {
+    this.builders.forEach(builder => builder.add(object))
+  }
+
+  build(ptr) {
+    this.builders.forEach(builder => builder.build(ptr))
+  }
+}
```
tpluscode: Surprised that eslint doesn't complain about multiple classes in one module. I think it's a good rule. 

Right now, for example, it was not obvious for me that there are multiple functionalities provided by the `Constraints.js` module.

I would propose to move each to a subdir, like `constrainBuilder/CompositeConstraintBuilder.js` etc and leave `Constraints.js` re-exporting them if needed. I would have the test file also split to avoid a very large test suite which I find hard to navigate

tpluscode (changes requested):
Let's start with a small change to the file organization.

Overall I very much like this approach.

I might have a general idea for a change, mapping the builders to the constraint they are building. For example, would we ever have multiple builders which populate `sh:nodeKind` for example? Else, if one were to replace an implementation it would be done by the constraint type (`sh:NodeKindConstraintComponent`, if memory serves)

Expanding on that, I think you could consider to refactor the composite builder in a ways that it produces `sh:or` as a combination of the inner builders?

packages/cube/lib/cube/buildCubeShape/Constraints.js
giacomociti: yes, according to [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_properties)

packages/cube/lib/cube/buildCubeShape/Constraints.js
giacomociti: Independent testability is one of the advantages of the approach proposed here, so splitting the test file definitively makes sense to me. But for the implementation classes I feel like breaking the 'one file per class' rule (as long as the classes are small and cohesive) because I tend to miss the big picture with too many classes scattered around. I know in enterprisey OO this rule is a must (and since I brought up OO patterns here I have to accept it :)).

giacomociti:
> consider to refactor the composite builder in a ways that it produces `sh:or` as a combination of the inner builders?

that's an interesting exercise but I already tried unsuccessfully. `DimensionConstraintsBuilder` (btw, I'd like to find a better name) is a kind of  *ad hoc* composite with specific business logic to combine constraints. I tried using a general-purpose `sh:or` composite builder but in some cases I was getting nested `sh:or` shapes (correct but cumbersome). Maybe we can make another attempt....

packages/cube/lib/cube/buildCubeShape/Constraints.js
tpluscode: Oh, that's not just syntactic sugar. I spend too much in TS only :)

tpluscode:
> Maybe we can make another attempt....

I trust that you tried already. Let's not spend time here if I'm likely missing a bigger picture or edge cases. Let me have another, more in-depth look at the details here

giacomociti:
I was thinking about adding more constraint generators, maybe for minLength, maxLength or languageIn.
Do you think they could be useful @tpluscode?

tpluscode:
Of course, but let's add new feature in future PRs

packages/cube/lib/cube/buildCubeShape/Cube.js
```diff
@@ -2,14 +2,19 @@ import cbdCopy from '../../cbdCopy.js'
 import Dimension from './Dimension.js'
 
 class Cube {
-  constructor({ rdf, metadata, observationSet, shape, term, propertyShapeId }) {
+  constructor({ rdf, metadata, observationSet, shape, term, propertyShapeId, inListThreshold }) {
```
tpluscode: Since this is going to be a breaking change for Cube Creator and something completely new here, maybe we can rename this option. In hindsight, I don't think this is descriptive enough.

`inListMaxSize` instead?

packages/cube/test/cube/constraintBuilder/support.js
```diff
@@ -0,0 +1,38 @@
...
+  const shape = rdf.clownface()
+  const ptr = shape.blankNode()
+  shape.namedNode('http://example.org/shape')
+    .addOut(sh.targetNode, subject)
+    .addOut(sh.property, ptr)
+  ptr.addOut(sh.path, property)
+
+  values.forEach(x => builder.add(x))
+  builder.build(ptr)
+  // console.log(shape.dataset.toCanonical())
```
tpluscode:
```suggestion
```

packages/cube/test/cube/constraintBuilder/DatatypeConstraintBuilder.test.js
```diff
@@ -0,0 +1,26 @@
...
+  const integer = rdf.literal('2', xsd.integer)
+  const namedNode = rdf.namedNode('http://example.org/namedNode')
+
+  it('should create a constrained shape', () => {
+    const builder = new DatatypeConstraintBuilder(rdf, xsd.string)
+    const validator = createValidator(builder, [string1, string2])
+    conforms(validator, [string3])
+    notConforms(validator, [integer])
+    notConforms(validator, [namedNode])
+  })
```
tpluscode: I'd prefer test cases separated. A method for many similar test cases is to loop over inputs of a similar test setup. Like below + some finetuning.

```js
[
  ['string', conforms, [string3]],
  ['integer', notConforms, [integer]],
  ['named node', notConforms, [namedNode]],
].forEach((when, assert, arg) =>
  context(when, () => 
    it('should create a constrained shape', () => {
      const builder = new DatatypeConstraintBuilder(rdf, xsd.string)
      const validator = createValidator(builder, [string1, string2])
      assert(validator, arg)
    })
  )
```

Else, if that's overkill, I would simply copy/paste test cases to isolate

packages/cube/test/cube/constraintBuilder/support.js
```diff
@@ -0,0 +1,38 @@
...
+  const report = validator.validate(createDataset(values))
+  strictEqual(report.conforms, true)
+}
+
+export const notConforms = (validator, values) => {
+  const report = validator.validate(createDataset(values))
+  strictEqual(report.conforms, false)
+}
+
+export const createValidator = (builder, values) => {
```
tpluscode: This threw me off at first. I know that technically it returns a validator but maybe a better name would be...

```suggestion
export const buildShape = (builder, values) => {
```

packages/cube/test/cube/constraintBuilder/DatatypeConstraintBuilder.test.js
tpluscode: Here's another idea, moving the test setup outside `it`. I think this would be useful in testing some other constraint builders (+curried assertion)

```js
context('built from all strings', () => {
  const builder = new DatatypeConstraintBuilder(rdf, xsd.string)
  const validator = createValidator(builder, [string1, string2])
  const assertConforms = conforms.bind(null, validator)
  const assertNotConforms = notConforms.bind(null, validator)

  it('string conforms', () => assertConforms([string3]))
  it('integer does not conform', () => assertNotConforms([integer]))
  it('named node does not conform', () => assertNotConforms([namedNode]))
})
```

packages/cube/test/cube/constraintBuilder/support.js
```diff
@@ -0,0 +1,38 @@
...
+
+export const conforms = (validator, values) => {
+  const report = validator.validate(createDataset(values))
+  strictEqual(report.conforms, true)
+}
+
+export const notConforms = (validator, values) => {
+  const report = validator.validate(createDataset(values))
+  strictEqual(report.conforms, false)
+}
```
tpluscode: Have you considered variadic arg for `values`? This would avoids arrays when calling in tests

```suggestion
export const conforms = (validator, ...values) => {
  const report = validator.validate(createDataset(values))
  strictEqual(report.conforms, true)
}

export const notConforms = (validator, ...values) => {
  const report = validator.validate(createDataset(values))
  strictEqual(report.conforms, false)
}
```

packages/cube/test/cube/constraintBuilder/ValuesConstraintBuilder.test.js
```diff
@@ -0,0 +1,28 @@
...
+    const validator = createValidator(builder, [integer, namedNode, blankNode])
+    conforms(validator, [string, integer, namedNode, blankNode])
+    notConforms(validator, [other])
+  })
+  it('should create an unconstrained shape due to too many values', () => {
+    const threshold = 2
+    const builder = new ValuesConstraintBuilder(rdf, string, threshold)
+    const validator = createValidator(builder, [integer, namedNode, blankNode])
+    conforms(validator, [string, integer, namedNode, blankNode, other])
+    strictEqual(builder.message, 'Too many values for in-list constraint.')
```
tpluscode: Are these messages used anywhere else but tests?

tpluscode (changes requested):

packages/cube/lib/cube/buildCubeShape/Cube.js
giacomociti: done

packages/cube/test/cube/constraintBuilder/support.js
giacomociti: done

packages/cube/test/cube/constraintBuilder/DatatypeConstraintBuilder.test.js
giacomociti: done this, very nice idea

packages/cube/test/cube/constraintBuilder/support.js
giacomociti: renamed

packages/cube/test/cube/constraintBuilder/support.js
giacomociti: done

packages/cube/test/cube/constraintBuilder/ValuesConstraintBuilder.test.js
giacomociti: not really.

tpluscode (approved):

PR #227, opened 2023-11-29
Prioritise explicitly set variables

tpluscode:
Currently, if a pipeline was started form a script like

```
export FOO=bar
npx barnard59 --variable-all --variable FOO=baz run ...
```

The value of `FOO` would be that from the environment. I find this behaviour a bug, thus a `patch` bump in the changset. On the other hand, existing pipelines may unwittingly rely on it. One might argue it would thus justify a `major` bump, again.

### DIFF
--- a/.changeset/good-readers-thank.md
+++ b/.changeset/good-readers-thank.md
@@ -1,0 +1,5 @@
+---
+"barnard59": patch
+---
+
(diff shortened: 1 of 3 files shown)


ludovicm67 (approved):
I would go for patch also, as it's a bug and that this scenario (having variables defined at multiple locations) was never documented.

PR #234, opened 2023-12-11
feat: type declarations in core package

tpluscode:
I'm a little tired of maintaining DTS package for barnard59 so I start with updating the core package to become TS

I kept all test files as plain JS because that greatly reduces the cost to migrate and does not pose any problems


packages/core/lib/Pipeline.ts
```diff
@@ -115,7 +147,7 @@ class Pipeline extends StreamObject {
 
   // stream interface
 
-  async _read(size) {
+  async _read(size: number) {
     try {
       await this.init()
```
giacomociti, line 152: not related to this PR, but my IDE suggests that `await` is not needed

packages/core/lib/Pipeline.ts
```diff
@@ -144,22 +176,27 @@ class Pipeline extends StreamObject {
...
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    } catch (err: any) {
       this._span.recordException(err)
       this.lastChild.stream.destroy(err)
     }
   }
 
-  async _write(chunk, encoding, callback) {
+  async _write(chunk: unknown, encoding: string, callback: (error?: Error | null) => void) {
     await this.init()
```
giacomociti, line 187: not related to this PR, but my IDE suggests that `await` is not needed

packages/core/lib/Pipeline.ts
```diff
@@ -144,22 +176,27 @@ class Pipeline extends StreamObject {
...
 
+    assertWritable(this.firstChild)
+
     this._span.addEvent('write')
     return this.firstChild.stream.write(chunk, encoding, callback)
   }
 
-  async _final(callback) {
+  async _final(callback: (error?: Error | null) => void) {
     await this.init()
```
giacomociti, line 196: not related to this PR, but my IDE suggests that `await` is not needed

ludovicm67 (approved):

packages/core/lib/Pipeline.ts
tpluscode: Actually, it was related as you explained. Thanks!

PR #241, opened 2023-12-21
Batch operation

giacomociti:
I think a batch operation can be useful and it belongs to the base package because it's a kind of inverse of `flatten`.

```turtle
@prefix p: <https://pipeline.described.at/> .
@prefix base: <https://barnard59.zazuko.com/operations/base/> .
@prefix n3: <https://barnard59.zazuko.com/operations/formats/n3/> .
@prefix ntriples: <https://barnard59.zazuko.com/operations/formats/ntriples/> .

<echo> a p:Pipeline , p:Readable;
  p:steps
    [
      p:stepList
      (
        [ base:stdin () ]
        [ n3:parse () ]
       
        [ base:batch (100) ]
        [ base:flatten () ]

        [ ntriples:serialize () ]
      )
    ]
.

```


.changeset/young-geese-teach.md
```diff
@@ -0,0 +1,6 @@
+---
+"barnard59-base": patch
+"barnard59-cube": patch
```
tpluscode, line 3: I would not include `-cube` in this changeset. It's going to receive a bump anyway, because it depends on `-base`

packages/base/batch.js
```diff
@@ -0,0 +1,26 @@
+import { Transform } from 'stream'
```
tpluscode: I wonder. Are we using `stream` or `readable-stream`? Does it make sense to mix?

That said, what about `through2`?

packages/cube/pipeline/check-observations.ttl
```diff
@@ -45,7 +45,8 @@ _:sortChunkSize a p:Variable ;
         [ n3:parse () ]
         _:sortBySubject
         [ splitDataset:bySubject () ]
-        _:batch
+        [ base:batch ("batchSize"^^p:VariableName) ]
+        [ base:flatten () ]
```
tpluscode, line 49: `flatten` was part of an experiment?

tpluscode (changes requested):

packages/base/batch.js
giacomociti: `through2` can be used as well, I used `Transform` after reading [this hint](https://www.npmjs.com/package/through2#do-you-need-this). But I'm not aware of the differences between `stream` and `readable-stream`.

packages/cube/pipeline/check-observations.ttl
giacomociti: here `flatten` is needed because chunks produced by `base:batch` are arrays of arrays (due to the preceding split by subject)

giacomociti: and in the previous `_:batch` implementation we were essentially doing the batching (`chunkObjectsBySize`) and the flattening (`chunkBySize`)

.changeset/young-geese-teach.md
tpluscode: Actually, I was wrong. Given the dependencies, `cube` will not be bumped because the new version `2.3` satisfies the range `^2.2.0`

We have 3 options (from bad to less bad):

1. Change the range to `~2.2.0`
2. Keep what you had before
3. Use experimental changesets option `updateInternalDependents: always`

The first is bad because we don't really want more strict range here. Sure path to duplicate resolutions of what should be compatible versions
Second is ok, but one needs to consider all dependencies every time. 
Last is an unstable feature but will ensure that any release of `b59-base`, for example, will also bump all packages which use it. This may be useful to ensure that we always get latest versions

.changeset/young-geese-teach.md
```diff
@@ -0,0 +1,5 @@
+---
+"barnard59-base": patch
```
tpluscode: That said, I also think this deserves a minor version. And upon consideration, I'd bring back cube bump

```suggestion
"barnard59-base": minor
"barnard59-cube": patch
```

packages/cube/pipeline/check-observations.ttl
tpluscode: Got it. Thanks

.changeset/young-geese-teach.md
tpluscode: Guess experiments do not deserve dedicated docs. Here it is in the changelog: https://github.com/changesets/changesets/blob/main/packages/cli/CHANGELOG.md#2160

tpluscode (approved):

PR #243, opened 2023-12-22
fix: resolving steps in global context

tpluscode:
Slightly hackish way so that barnard packages installed globally can be used using. Tweaks were needed to support manifest resolution and importing steps from pipelines (not limited to shorthand syntax)


packages/cli/lib/discoverManifests.js
```diff
@@ -35,14 +36,18 @@ export default async function * () {
   }
 }
 
-async function getInstalledPackages() {
+async function getInstalledPackages(all) {
   if (isInstalledGlobally) {
+    let npmList = 'npm list -g'
+    if (all) {
+      npmList += ' --all'
+    }
```
tpluscode, line 44: When resolving steps for shorthand syntax, we need all packages, not inly those installed directly

packages/cli/lib/discoverManifests.js
```diff
@@ -9,10 +8,12 @@ import { packageDirectory } from 'pkg-dir'
 import { findUp } from 'find-up'
 
 const packagePattern = /^barnard59-(.+)$/
-const require = module.createRequire(import.meta.url)
 
-export default async function * () {
-  const packages = (await getInstalledPackages()).filter(hasManifest)
+export default async function * ({ basePath = import.meta.url, all = false } = {}) {
+  const require = module.createRequire(basePath)
+  const hasManifest = canRequireManifest.bind(null, require)
```
tpluscode, line 14: The `basePath` cannot be the module URL because in global context, `barnard59-*` packages cannot be resolved from the CLI package. Thus, when loading the pipeline, its path will be used as base for resolution

packages/cli/lib/pipeline.js
```diff
@@ -20,14 +25,19 @@ const discoverOperations = async () => {
...
 }
 
-export const desugar = async (dataset, { logger, knownOperations } = {}) => {
-  knownOperations = knownOperations ?? await discoverOperations()
+export const desugar = async (dataset, { logger, knownOperations, pipelinePath } = {}) => {
+  knownOperations = knownOperations ?? await discoverOperations(pipelinePath)
+  const dir = await packageDirectory({
+    cwd: pipelinePath,
+  })
+  const require = module.createRequire(dir + '/')
```
tpluscode, line 33: Somehow, the base dir to require must be the path, ending in `/`. This is only important in global context

packages/cli/lib/pipeline.js
```diff
@@ -38,7 +48,17 @@ export const desugar = async (dataset, { logger, knownOperations } = {}) => {
...
+      let { type, link } = knownStep
+      const { pathname, filename, hash, protocol } = iriResolve(link.value, pipelinePath)
+
+      if (protocol === 'node:' && isInstalledGlobally) {
+        // must resolve node: links to absolute paths to overcome node resolution issues in global context
+        const resolved = require.resolve(pathname)
+        if (filename !== resolved) {
+          link = rdf.namedNode(url.pathToFileURL(resolved).toString() + hash)
+        }
+      }
```
tpluscode, line 60: And then, in global context, `node:` implementations of steps must be replaced with absolute `file:` URLs.

Somehow this cannot be done uniformly in local installation...

PR #247, opened 2023-12-29
Generators

tpluscode:


giacomociti:
I like generators so I'm really happy about this PR. I tried to understand if and when there are performance penalties compared to the lower level streams APIs. For example, streams created `from` a generator have `highWaterMark` fixed to `1` and I'm not sure if this can cause some kind of 'impedance mismatch' with other streams. 

One example in which a generator is considerably slower is the following:
```js
import { Duplex, Transform } from 'node:stream'

const transform1 = new Transform({
  readableObjectMode: true,
  writableObjectMode: true,
  transform(chunk, _encoding, callback) {
    callback(null, chunk)
  },
})

const transform2 = Duplex.from(async function * (stream) {
  for await (const chunk of stream) {
    yield chunk
  }
})
```

The two transformations above have the same dummy behaviour but perform differently (I don't know why):

```js
import rdfUtils from 'rdf-utils-fs'

async function run(transform) {
  const input = rdfUtils.fromFile('input.ttl')
  await rdfUtils.toFile(input.pipe(transform), 'output.ttl')
}

await run(transform1) // fast
await run(transform2) // slow

```

Maybe it's just something fishy with `rdfUtils.toFile` because the perf issue disappears with the following code:

```js
import { pipeline } from 'node:stream/promises'
import { ntriplesSerialize } from 'barnard59-formats'
import fs from 'fs'

async function run(transform) {
  const input = rdfUtils.fromFile(inputFile)
  const output = fs.createWriteStream(outputFile)
  await pipeline(input, ntriplesSerialize(), output)
}
```
and fortunately the latter is what we usually have in barnard59 pipelines.

tpluscode:
Thank you for the analysis. I found this related SO question: https://stackoverflow.com/a/71547827/1103498

From what I understand, `highWaterMark` affect how much gets buffered when the stream reads. Generator always works on chunk at a time so that's why it's hardcoded to `1`. It is quite curious why `toFile` behaves like that.

test/e2e/definitions/limit-offset.ttl
```diff
@@ -0,0 +1,15 @@
...
+@prefix p: <https://pipeline.described.at/> .
+@prefix code: <https://code.described.at/> .
+@prefix base: <https://barnard59.zazuko.com/operations/base/> .
+
+<> a p:Pipeline, p:ReadableObjectMode ;
+  p:steps
+    [
+      p:stepList
+        (
+          [ base:streamValues ( "{ age: 34 }"^^code:EcmaScript "{ age: 23 }"^^code:EcmaScript "{ age: 10 }"^^code:EcmaScript) ]
```
tpluscode, line 11: What do you think about the identifier `base:streamValues`? @giacomociti

giacomociti: I like it. Even though the name of the function implementing it is completely different, I think it makes sense to use a higher-level language in the pipeline

packages/base/test/Readable.test.js
```diff
@@ -0,0 +1,13 @@
...
+
+describe('Readable', () => {
+  it('should return a readable stream', () => {
+    // given
+    const stream = Readable.from([0, 1, 2, 3, 4])
+
+    // then
+    assert(isReadableStream(stream))
+  })
+})
```
giacomociti: we should leverage the spread syntax:
```suggestion
import assert from 'node:assert'
import { isReadableStream } from 'is-stream'
import getStream from 'get-stream'
import { expect } from 'chai'
import * as Readable from '../Readable.js'

describe('Readable', () => {
  it('should return a readable stream', async () => {
    // given
    const stream = Readable.from(0, 1, 2, 3, 4)

    // then
    assert(isReadableStream(stream))
    const array = await getStream.array(stream)
    expect(array).to.deep.equal([0, 1, 2, 3, 4])
  })
})
```

giacomociti (approved):

packages/base/test/Readable.test.js
giacomociti, line 1: ops, I meant to suggest the change, not to directly commit. Hope it is fine

tpluscode: Yes, perfect :)

PR #254, opened 2024-01-08
fix: b59 run would fail on relative pipeline path

tpluscode:
I added a CI job which runs the test e2e pipelines from the b59 CLI. This revealed the regression where pipeline path was no longer properly resolved against working dir

Also need to update some test infrastructure

### DIFF
--- a/packages/cli/lib/pipeline.js
+++ b/packages/cli/lib/pipeline.js
@@ -109,7 +109,7 @@ async function fileToDataset(filename) {
  */
 export async function parse(filename, iri, { logger } = {}) {
   const dataset = await fileToDataset(filename)
-  const ptr = findPipeline(await desugar(dataset, { logger, pipelinePath: filename }), iri)
+  const ptr = findPipeline(await desugar(dataset, { logger, pipelinePath: resolve(filename) }), iri)
 
   return {
     basePath: resolve(dirname(filename)),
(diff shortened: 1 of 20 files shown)


packages/cli/lib/pipeline.js
```diff
@@ -109,7 +109,7 @@ async function fileToDataset(filename) {
  */
 export async function parse(filename, iri, { logger } = {}) {
   const dataset = await fileToDataset(filename)
-  const ptr = findPipeline(await desugar(dataset, { logger, pipelinePath: filename }), iri)
+  const ptr = findPipeline(await desugar(dataset, { logger, pipelinePath: resolve(filename) }), iri)
```
tpluscode, line 112: Te fix is a mere one-liner :)

test/e2e/operations/iterateDirectory.js
```diff
@@ -14,6 +14,11 @@ class FileIterator extends Readable {
     const directory = resolve(basePath, dirName)
 
     readdir(directory, (e, files) => {
+      if (e) {
```
giacomociti, line 17: what is the purpose of this?

giacomociti (approved):

test/e2e/operations/iterateDirectory.js
tpluscode: That's just for tests. Without forwarding the error, it was swallowed

PR #263, opened 2024-01-18
More cube fetch

giacomociti:


packages/cube/lib/report.js
```diff
@@ -0,0 +1,29 @@
+import termToNt from '@rdfjs/to-ntriples'
+import rdf from 'barnard59-env'
```
tpluscode: Please avoid using this directly and get from `this.env` in the step factory

packages/cube/package.json
```diff
@@ -20,24 +20,25 @@
   },
   "homepage": "https://github.com/zazuko/barnard59",
   "dependencies": {
+    "barnard59-env": "^1.2.0",
```
tpluscode: And then move back to dev deps

packages/cube/manifest.ttl
```diff
@@ -48,3 +48,24 @@
...
+  rdfs:label "Retrieves cube with observations from SPARQL endpoint" ;
+  b59:source "barnard59-cube/pipeline/fetch-cube.ttl" ;
+.
+
+<command/cube/report-summary>
+  a b59:CliCommand ;
+  b59:command "report-summary" ;
+  rdfs:label "Human-readable summary of SHACL validation report" ;
+  b59:source "barnard59-cube/pipeline/report-summary.ttl" ;
+.
```
tpluscode: I though this belonged to the `b59-shacl` package. Then adding dependency on `rdf-validate-shacl` would not be necessary in this `b59-cube`

tpluscode (changes requested):

packages/cube/lib/report.js
giacomociti: done, I was not aware that [`base:map`](https://github.com/zazuko/barnard59/blob/2c64f4da790ceab0b4dd1795a81e55de4cfdacc6/packages/base/map.js#L45) is smart enough to inject the context

tpluscode: Yes, we added that a some point so that you can access variables

tpluscode (approved):
I took the liberty of adding a second changeset

PR #264, opened 2024-01-19
build: @zazuko/env v2

tpluscode:


packages/env/index.ts
```diff
@@ -1,8 +1,8 @@
-import { create } from '@zazuko/env-node'
-import type { DefaultEnv, DerivedEnvironment } from '@zazuko/env'
+import BaseEnv, { Environment as ZazukoEnv } from '@zazuko/env-node'
 import NamespacesFactory from './lib/Namespaces.js'
 import ConstantsFactory from './lib/Constants.js'
 
-export type Environment = DerivedEnvironment<DefaultEnv, ConstantsFactory>
+const env = new ZazukoEnv([NamespacesFactory, ConstantsFactory], { parent: BaseEnv })
```
tpluscode, line 5: This is the core of the PR. The dev-friendly method for combining multiple environments

Rest is updating dependencies to ensure we only have the latest libs, without duplicates

giacomociti: thanks for the tip

giacomociti (approved):

PR #273, opened 2024-01-30
Shacl report success

giacomociti:
as noted here: https://github.com/zazuko/cube-link/pull/128#issuecomment-1915021022

### DIFF
--- a/packages/shacl/lib/TermCounter.js
+++ b/packages/shacl/lib/TermCounter.js
@@ -1,8 +1,14 @@
 export default class TermCounter {
+  /**
+   * @param {import('@rdfjs/term-map/Factory.js').TermMapFactory} env
+   */
   constructor(env) {
     this.termMap = env.termMap()
   }
 
+  /**
+   * @param {import('@rdfjs/types').Term} term
+   */
   add(term) {
     this.termMap.set(term, (this.termMap.get(term) ?? 0) + 1)
   }
--- a/packages/shacl/lib/TermCounter.js
+++ b/packages/shacl/lib/TermCounter.js
@@ -1,0 +1,9 @@
+export default class TermCounter {
+  constructor(env) {
+    this.termMap = env.termMap()
+  }
+
+  add(term) {
+    this.termMap.set(term, (this.termMap.get(term) ?? 0) + 1)
+  }
+}
--- a/.changeset/lemon-icons-clean.md
+++ b/.changeset/lemon-icons-clean.md
@@ -1,0 +1,5 @@
+---
+"barnard59-shacl": patch
+---
+
+Produce SHACL report on successful validation
--- a/packages/cube/test/validation.pipeline.test.js
+++ b/packages/cube/test/validation.pipeline.test.js
@@ -15,7 +15,7 @@ describe('cube validation pipeline', function () {
     const result = shell.exec(command, { silent: true, cwd })
 
     strictEqual(result.stderr, '')
-    strictEqual(result.stdout, '')
+    ok(result.stdout.includes('_:report <http://www.w3.org/ns/shacl#conforms> "true"^^<http://www.w3.org/2001/XMLSchema#boolean>'))
     strictEqual(result.code, 0)
   })
 
@@ -48,7 +48,7 @@ describe('cube validation pipeline', function () {
     const result = shell.exec(command, { silent: true, cwd })
 
     strictEqual(result.stderr, '')
-    strictEqual(result.stdout, '')
+    ok(result.stdout.includes('_:report <http://www.w3.org/ns/shacl#conforms> "true"^^<http://www.w3.org/2001/XMLSchema#boolean>'))
     strictEqual(result.code, 0)
   })
 })
--- a/packages/shacl/lib/report.js
+++ b/packages/shacl/lib/report.js
@@ -23,6 +23,16 @@ function getMessages(report) {
     .map(message => message + '\n')
 }
 
-export function getSummary(dataset) {
-  return getMessages(new ValidationReport(this.env.clownface({ dataset })))
+export function getSummary() {
+  return async function * (stream) {
+    let total = 0
+    for await (const dataset of stream) {
+      const messages = getMessages(new ValidationReport(this.env.clownface({ dataset })))
+      total += messages.length
+      yield messages
+    }
+    if (total) {
+      this.error(new Error(`${total} violations found`))
+    }
+  }.bind(this)
 }
--- a/packages/shacl/lib/report.js
+++ b/packages/shacl/lib/report.js
@@ -33,6 +33,8 @@ export function getSummary() {
     }
     if (total) {
       this.error(new Error(`${total} violations found`))
+    } else {
+      yield 'successful\n'
     }
   }.bind(this)
 }
--- a/packages/shacl/pipeline/report-summary.ttl
+++ b/packages/shacl/pipeline/report-summary.ttl
@@ -13,10 +13,10 @@
         [ base:stdin () ]
         [ n3:parse () ]
         [ rdf:getDataset () ]
-        [ base:map ([ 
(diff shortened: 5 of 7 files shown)


tpluscode (approved):
It was intentional to have no output but I suppsoe we can have it is requested

Rdataflow:
@giacomociti thank you 😄

Rdataflow:
@giacomociti can you add returnvalue > 0 on failure?

currently the validation fails but inconsistently returns 0:
```
cat cube.ttl | npx barnard59 cube check-observations --constraint constraint.ttl | npx barnard59 shacl report-summary && echo SUCCESS || echo FAIL

error: 13 violations found {"stack":"Error: 13 violations found\n    at Object.validate (file:///tmp/cube-link/node_modules/barnard59-shacl/report.js:31:16)\n    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)\n    at async nextAsync (node:internal/streams/from:182:33)","timestamp":"2024-01-30T19:07:25.899Z"}
Error: 13 violations found
    at Object.validate (file:///tmp/cube-link/node_modules/barnard59-shacl/report.js:31:16)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async nextAsync (node:internal/streams/from:182:33) {
  level: 'error',
  timestamp: '2024-01-30T19:07:25.899Z',
  [Symbol(level)]: 'error'
}
Violation of ClassConstraintComponent: "" with path <http://ns.bergnet.org/dark-horse#room> at focus node <http://example.org/cube-link-example/temperature-sensor/cube/observation/20190102T120000058Z> (source: _:b87_b606)
Violation of ClassConstraintComponent: "" with path <http://ns.bergnet.org/dark-horse#room> at focus node <http://example.org/cube-link-example/temperature-sensor/cube/observation/20190103T120000055Z> (source: _:b87_b606)
Violation of ClassConstraintComponent: "" with path <http://ns.bergnet.org/dark-horse#room> at focus node <http://example.org/cube-link-example/temperature-sensor/cube/observation/20190104T120000013Z> (source: _:b87_b606)
Violation of ClassConstraintComponent: "" with path <http://ns.bergnet.org/dark-horse#room> at focus node <http://example.org/cube-link-example/temperature-sensor/cube/observation/20190105T120000041Z> (source: _:b87_b606)
Violation of ClassConstraintComponent: "" with path <http://ns.bergnet.org/dark-horse#room> at focus node <http://example.org/cube-link-example/temperature-sensor/cube/observation/20190106T120000057Z> (source: _:b87_b606)
Violation of ClassConstraintComponent: "" with path <http://ns.bergnet.org/dark-horse#room> at focus node <http://example.org/cube-link-example/temperature-sensor/cube/observation/20190107T120000059Z> (source: _:b87_b606)
Violation of ClassConstraintComponent: "" with path <http://ns.bergnet.org/dark-horse#room> at focus node <http://example.org/cube-link-example/temperature-sensor/cube/observation/20190108T120000017Z> (source: _:b87_b606)
Violation of ClassConstraintComponent: "" with path <http://ns.bergnet.org/dark-horse#room> at focus node <http://example.org/cube-link-example/temperature-sensor/cube/observation/20190109T120000055Z> (source: _:b87_b606)
Violation of ClassConstraintComponent: "" with path <http://ns.bergnet.org/dark-horse#room> at focus node <http://example.org/cube-link-example/temperature-sensor/cube/observation/20190110T120000059Z> (source: _:b87_b606)
Violation of ClassConstraintComponent: "" with path <http://ns.bergnet.org/dark-horse#room> at focus node <http://example.org/cube-link-example/temperature-sensor/cube/observation/20190111T120000055Z> (source: _:b87_b606)
Violation of ClassConstraintComponent: "" with path <http://ns.bergnet.org/dark-horse#room> at focus node <http://example.org/cube-link-example/temperature-sensor/cube/observation/20190112T120000055Z> (source: _:b87_b606)
Violation of ClassConstraintComponent: "" with path <http://ns.bergnet.org/dark-horse#room> at focus node <http://example.org/cube-link-example/temperature-sensor/cube/observation/20190113T120000055Z> (source: _:b87_b606)
Violation of ClassConstraintComponent: "" with path <http://ns.bergnet.org/dark-horse#room> at focus node <http://example.org/cube-link-example/temperature-sensor/cube/observation/20190114T120000015Z> (source: _:b87_b606)
SUCCESS
```

tpluscode:
Ah, @giacomociti, what will the `shacl report-summary` command do when it gets a success report?

giacomociti:
> Ah, @giacomociti, what will the `shacl report-summary` command do when it gets a success report?

nothing, empty stdout

giacomociti:
> @giacomociti can you add returnvalue > 0 on failure?

@Rdataflow you mean the return value of the `report-summary` command? Because the `check-observation` command should already return 1 in case of validation failure. What do you think @tpluscode? I understand in practice this is definitively useful, the only perplexity is that we would have a failure of `report-summary` when the command perform its duty without any issue. How to distinguish with real issues due to bugs or unexpected input?

tpluscode:
Good point. I think we'd need to support both scenarios with a flag. Principle to least astonishment would suggest that we should by default fail when the report is `sh:conforms false`. Thus, I'd propose to add option to ignore conformance
   
```
b59 shacl report-summary --ignore-sh-conforms
```

Better name for this flag appreciated.

By the way, we currently don't have a way to declare boolean options. Would need to add this to the CLI

Rdataflow:
> > @giacomociti can you add returnvalue > 0 on failure?
> 
> @Rdataflow you mean the return value of the `report-summary` command? Because the `check-observation` command should already return 1 in case of validation failure. What do you think @tpluscode? I understand in practice this is definitively useful, the only perplexity is that we would have a failure of `report-summary` when the command perform its duty without any issue. How to distinguish with real issues due to bugs or unexpected input?

@giacomociti exactly I was thinking of the `report-summary` command which is last in the pipe
`cat cube.ttl | npx barnard59 cube check-observations --constraint constraint.ttl | npx barnard59 shacl report-summary && echo SUCCESS || echo FAIL`

Rdataflow:
> > Ah, @giacomociti, what will the `shacl report-summary` command do when it gets a success report?
> 
> nothing, empty stdout

@giacomociti you may add a one-liner 
i.e. similar to what you have in https://github.com/zazuko/cube-link/blob/a722faa42ac87273deceff62f1ff198e136c5311/bin/cube-link.js#L39

packages/shacl/lib/TermCounter.js
```diff
@@ -0,0 +1,15 @@
...
+  constructor(env) {
+    this.termMap = env.termMap()
+  }
+
+  /**
+   * @param {import('@rdfjs/types').Term} term
+   */
+  add(term) {
+    this.termMap.set(term, (this.termMap.get(term) ?? 0) + 1)
+  }
```
tpluscode, line 14: Technically, the result should have a severity but the `rdf-validate-shacl` has it correct, that it may be missing. To accommodate for that case, I would allow `null` here and add a separate counter for a total

```suggestion
  constructor(env) {
    this.total = 0
    this.termMap = env.termMap()
  }

  /**
   * @param {import('@rdfjs/types').Term | null} term
   */
  add(term) {
    this.total += 1
    if (term) {  
      this.termMap.set(term, (this.termMap.get(term) ?? 0) + 1)
    }
  }
```

tpluscode (changes requested):

tpluscode (approved):

packages/shacl/lib/TermCounter.js
giacomociti: initially I allowed `null` but later on I changed my mind: handling a case with no business value is accidental complexity.
The downside is I had to move the `if` in the calling code. There, I would prefer to assert that severity is not null (and throw in case the report is not valid) but I don't know how to do it in JavaScript (Non-Null Assertion Operator is TypeScript only)

PR #275, opened 2024-02-08
fix: constraint missing from `fetch-cube` output

tpluscode:


giacomociti:
The initial idea was to provide the same kind of data as in the existing cube.link [example](https://github.com/zazuko/cube-link/blob/main/cube.ttl): everything except the constraint. Since the constraint part is not expected to be huge, it makes sense to include it.

tpluscode:
Yes, it should be fairly small in relation to the all the observation data.

The background is that, given a small cube, it should be enough to just fetch the whole thing and use for all validations

giacomociti (approved):

PR #282, opened 2024-03-18
sparql-http-client v3

tpluscode:


tpluscode:
@giacomociti the last failing tests are `put/post should do nothing if the stream was closed and no quads have been written`

Something must have changed in the library that a request is being sent anyway. I don't know if this OK and we can safely remove these tests or if it is an important regression 🤔

giacomociti:
the tests were introduced in commit https://github.com/zazuko/barnard59/commit/c49bea0fc17a532b15c1cc7d56edabaa7fe27b2c but I'm not able to relate them to the commit message "fix: close writable if no quads have been written"

tpluscode:
Looks like the test was added to simply match the implemented behaviour.

Guess it's a consequence of the stream splitter being removed. Previous version created a new request for subsequent graphs in data stream. No request if you had not a single quad to decide where to write to. Now it simply sends whatever comes its way.

On second thought, an empty stream is something that could be a legitimate output of a pipeline. `POST` would do nothing and `PUT` would clear the target graph?

`graph-store` is already a breaking release in this PR so might just go with what we have? Add a warning to the log if `POST`/`PUT` happens without any data?

giacomociti:
in pipeline `put.ttl` the step `op:graph-store\/put` needs one more input argument:

```turtle
[
      code:name "graph" ;
      code:value "graph"^^p:VariableName ;
],
```

tpluscode:
> in pipeline `put.ttl` the step `op:graph-store\/put` needs one more input argument:
> 

will add an e2e test pipeline!

giacomociti (approved):

Issue #304, opened 2024-05-20
Differences between the toCubeShape used in Cube Creator vs b59-cube


tpluscode:
As discussed, in all cases the `sh:in` gets separated between the `sh:or` alternative shapes.

This would make it difficult for consumers to analyse the cube contents

In the cases when there are multiple `sh:in` lists, they should be merged and appear in the property shape itself.

### Dimension with `""^^:cube:Undefined`

```diff
--- before.ttl	2024-05-20 10:31:34
+++ after.ttl	2024-05-20 10:32:16
@@ -2,17 +2,19 @@
     sh:or (
               [
                   sh:datatype xsd:string ;
+                  sh:in (
+                            "incomplete series"
+                        ) ;
               ]
               [
                   sh:datatype <https://cube.link/Undefined> ;
+                  sh:in (
+                            ""^^<https://cube.link/Undefined>
+                        ) ;
               ]
           ) ;
-    sh:in (
-              "incomplete series"
-              ""^^<https://cube.link/Undefined>
-          ) ;
     sh:maxCount 1 ;
     sh:minCount 1 ;
     sh:nodeKind sh:Literal ;
     sh:path rdfs:comment ;
 ]
```

### Numeric dimension with `""^^cube:Undefined` literals

```diff
--- before.ttl	2024-05-20 10:40:23
+++ after.ttl	2024-05-20 10:39:30
@@ -2,15 +2,18 @@
     sh:or (
               [
                   sh:datatype <https://cube.link/Undefined> ;
+                  sh:in (
+                            ""^^<https://cube.link/Undefined>
+                        ) ;
               ]
               [
+                  sh:maxInclusive 500 ;
+                  sh:minInclusive "1"^^xsd:int ;
                   sh:datatype xsd:int ;
               ]
           ) ;
-    sh:maxInclusive "500"^^xsd:int ;
-    sh:minInclusive "1"^^xsd:int ;
     sh:maxCount 1 ;
     sh:minCount 1 ;
     sh:nodeKind sh:Literal ;
     sh:path <https://environment.ld.admin.ch/foen/ubd/28/dimension/limitvalue> ;
 ]
```

### Dimension with mixed IRIs and Literals

```diff
--- before.ttl	2024-05-20 10:37:26
+++ after.ttl	2024-05-20 10:37:41
@@ -1,19 +1,28 @@
 [
-    sh:in (
-              <https://cube.link/Undefined>
-              unit:MicroGM-PER-M3
-              "count"
[...]

PR #305, opened 2024-05-20
Fix 304

giacomociti:
should address #304.  As discussed, the main idea is, in case there is a `sh:or` with all the alternatives having their own `sh:in`, to merge all the values into a single, top-level `sh:in` constraint.

### DIFF
--- a/packages/cube/lib/cube/buildCubeShape/constraintBuilder/DimensionConstraintsBuilder.js
+++ b/packages/cube/lib/cube/buildCubeShape/constraintBuilder/DimensionConstraintsBuilder.js
@@ -4,6 +4,12 @@ import { RangeConstraintBuilder } from './RangeConstraintBuilder.js'
 import { ValuesConstraintBuilder } from './ValuesConstraintBuilder.js'
 import { NodeKindConstraintBuilder } from './NodeKindConstraintBuilder.js'
 
+const hasValuesBuilder = compositeBuilder =>
+  compositeBuilder.builders.some(builder => builder instanceof ValuesConstraintBuilder && builder.enabled)
+
+const getValuesBuilder = compositeBuilder =>
+  compositeBuilder.builders.find(builder => builder instanceof ValuesConstraintBuilder)
+
 export class DimensionConstraintsBuilder {
   constructor({ rdf, datatypeParsers, inListMaxSize }) {
     this.rdf = rdf
@@ -63,6 +69,17 @@ export class DimensionConstraintsBuilder {
       builders[0].build(ptr)
     }
     if (builders.length > 1) {
+      // if all builders have sh:in, then merge them
+      if (builders.every(hasValuesBuilder)) {
+        const merged = new ValuesConstraintBuilder(this.rdf)
+        for (const compositeBuilder of builders) {
+          const valuesBuilder = getValuesBuilder(compositeBuilder)
+          valuesBuilder.enabled = false
+          valuesBuilder.values.forEach(value => merged.add(value))
+        }
+        merged.build(ptr)
+      }
+
       ptr.addList(this.rdf.ns.sh.or, builders.map(builder => {
         const blankNode = ptr.blankNode()
         builder.build(blankNode)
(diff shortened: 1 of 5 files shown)


packages/cube/test/cube/constraintBuilder/DimensionConstraintsBuilder.test.js
```diff
@@ -84,4 +88,23 @@ describe('DimensionConstraintsBuilder', () => {
...
+  context('built from a string and a cube:Undefined', () => {
+    const builder = new DimensionConstraintsBuilder({ rdf, datatypeParsers })
+    const validator = buildShape(builder, string1, cubeUndefined)
+    const assertConforms = conforms.bind(null, validator)
+    const assertNotConforms = notConforms.bind(null, validator)
+
+    it('the two values conform', () => assertConforms(string1, cubeUndefined))
+    it('another string does not conform', () => assertNotConforms(string2))
+  })
+  context('built from a string and a named node', () => {
```
tpluscode: Side note: I think the test file will be more readable if you'd add separation between `context` and "long" `it`

```suggestion
    it('the two values conform', () => assertConforms(string1, cubeUndefined))
    it('another string does not conform', () => assertNotConforms(string2))
  })

  context('built from a string and a named node', () => {
```

packages/cube/lib/cube/buildCubeShape/constraintBuilder/ValuesConstraintBuilder.js
```diff
@@ -17,6 +17,10 @@ export class ValuesConstraintBuilder {
   }
 
   build(ptr) {
+    if (this.message) {
+      ptr.addOut(this.sh.description, this.message)
+      return
+    }
```
tpluscode, line 23: What is this message?

packages/cube/lib/cube/buildCubeShape/constraintBuilder/DimensionConstraintsBuilder.js
```diff
@@ -65,6 +69,17 @@ export class DimensionConstraintsBuilder {
...
+      // if all builders have sh:in, then merge them
+      if (builders.every(hasValuesBuilder)) {
+        const merged = new ValuesConstraintBuilder(this.rdf)
+        for (const compositeBuilder of builders) {
+          const valuesBuilder = getValuesBuilder(compositeBuilder)
+          valuesBuilder.enabled = false
+          valuesBuilder.values.forEach(value => merged.add(value))
+        }
+        merged.build(ptr)
+      }
```
tpluscode: Maybe it would be a little more readable if you'd simply loop the builders and `break` when any one of them does not have values builder

Trades `hasValuesBuilder` for `disableValuesBuilder` and reduces nesting. Up to you

```suggestion
      let merged = new ValuesConstraintBuilder(this.rdf)
      for (const compositeBuilder of builders) {
        const valuesBuilder = getValuesBuilder(compositeBuilder)
        if (!valuesBuilder) {
          merged = null
          break
        }

        valuesBuilder.values.forEach(value => merged.add(value))
      }
      if (merged) {
        builders.forEach(disableValuesBuilder)
        merged.build(ptr)
      }
```

tpluscode (approved):

packages/cube/lib/cube/buildCubeShape/constraintBuilder/ValuesConstraintBuilder.js
giacomociti: it is the fixed text 'Too many values for in-list constraint.', defined a few lines above. I think we discussed ways to make it more informative but decided to leave it like this for now.

packages/cube/lib/cube/buildCubeShape/constraintBuilder/DimensionConstraintsBuilder.js
giacomociti: what about instead:

```js
      const valuesBuilders = builders.map(getValuesBuilder)
      if (valuesBuilders.every(builder => builder && builder.enabled)) {
        const merged = new ValuesConstraintBuilder(this.rdf)
        const allValues = valuesBuilders.flatMap(builder => Array.from(builder.values))
        allValues.forEach(value => merged.add(value))
        valuesBuilders.forEach(builder => { builder.enabled = false })
        merged.build(ptr)
      }
```

tpluscode: Compact! :+1:

tpluscode: Does this work?

```js
if (valuesBuilders.every(builder => builder?.enabled)) {
```

PR #322, opened 2024-08-21
feat(shacl): validate command: look for shapes in validated chunk

tpluscode:
I wanted to be able to validate from the CLI command without `--shapes`. In this case, the standard input would be expected to include the shapes too

```
b59 shacl validate < data+shapes.ttl
```

As consequence, in any pipeline where we use the SHACL step, we can process each chunk individually, where each can have different shapes too.

### DIFF
--- a/packages/shacl/report.js
+++ b/packages/shacl/report.js
@@ -65,7 +65,7 @@ export async function shacl(arg) {
 
   let ds
   if (!shape) {
-    this.logger.info('No shapes found. Will validate each chunk against shapes found in teh chunk itself')
+    this.logger.info('No shapes found. Will validate each chunk against shapes found in the chunk itself')
   } else {
     if (!isReadableStream(shape)) {
       throw new Error(`${shape} is not a readable stream`)
--- a/packages/shacl/report.js
+++ b/packages/shacl/report.js
@@ -5,11 +5,12 @@ import TermCounter from './lib/TermCounter.js'
 
 /**
  * @this {import('barnard59-core').Context}
- * @param {import('@rdfjs/types').DatasetCore} ds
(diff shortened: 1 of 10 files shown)


packages/shacl/report.js
```diff
@@ -63,14 +63,18 @@ export async function shacl(arg) {
     maxViolations = options.maxErrors < 1 ? 0 : Number(options.maxErrors)
   }
 
+  let ds
   if (!shape) {
-    throw new Error('Needs a SHACL shape as parameter')
-  }
-  if (!isReadableStream(shape)) {
-    throw new Error(`${shape} is not a readable stream`)
+    this.logger.info('No shapes found. Will validate each chunk against shapes found in teh chunk itself')
```
giacomociti: typo `teh`

packages/shacl/report.js
```diff
@@ -19,8 +20,7 @@ async function * validate(ds, maxViolations, iterable) {
       break
     }
 
-    // create a new validator instance at each iteration to avoid memory leaks
-    const validator = new SHACLValidator(ds, { maxErrors: 0, factory: this.env })
+    const validator = new SHACLValidator(shapes || chunk, { maxErrors: 0, factory: this.env })
```
giacomociti, line 23: wondering about merging `shapes`  and `chunk`, but probably not worth it for the overhead and the added complexity

giacomociti (approved):

packages/shacl/report.js
tpluscode: Merging how?

giacomociti: regular graph merge, hence collecting shapes from both

PR #326, opened 2024-09-12
feat(map,filter): additional arguments

tpluscode:
I extended the `map` and `filter` steps with the option to pass any number of additional parameters for the callbacks.

Until now, to parametrise them, one required to use and access the variables. It should now be easier to achieve the same result.

For example, to filter only chunks which are greater than X

```turtle
[
  base:filter (
    "(chunk, _, minValue) => chunk > minValue"^^code:EcmaScript ;
    "X"^^p:VariableName ;
  ) 
] .
```

### DIFF
--- a/packages/base/test/map.test.js
+++ b/packages/base/test/map.test.js
@@ -25,6 +25,40 @@ describe('map', () => {
     deepStrictEqual(output, ['foo_a', 'foo_b'])
   })
 
+  it('should pass optional arguments to mapper function', async () => {
+    const input = new Readable({
+      read: () => {
+        input.push('a')
+        input.push('b')
+        input.push(null)
+      },
+    })
+
+    const prependPrefix = (chunk, prefix) => prefix + chunk
+    const outStream = input.pipe(map.call(context, prependPrefix, 'foo_'))
+    const output = await array(outStream)
+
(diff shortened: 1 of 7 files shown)


packages/base/test/map.test.js
```diff
@@ -25,6 +25,40 @@ describe('map', () => {
...
+  it('should pass optional arguments to mapper function', async () => {
+    const input = new Readable({
+      read: () => {
+        input.push('a')
+        input.push('b')
+        input.push(null)
+      },
+    })
+
+    const prependPrefix = (chunk, prefix) => prefix + chunk
```
giacomociti, line 37: the additional variadic argument is not needed using [partial application](https://en.wikipedia.org/wiki/Partial_application):
```js
const prependPrefix = prefix => chunk => prefix + chunk
const outStream = input.pipe(map.call(context, prependPrefix('foo_')))
```
But expressing partial application as code snippets b59 turtle files is more cumbersome, so probably the variadic arg solution is the best option

giacomociti (approved):

packages/base/test/map.test.js
tpluscode: This is an intriguing idea but I agree that the turtle representation would likely be more complicated

PR #327, opened 2024-09-12
ci: run e2e tests

tpluscode:
I noticed these were failing locally but not in the CI... because they were never executed?

### DIFF
--- a/packages/base/limit.js
+++ b/packages/base/limit.js
@@ -12,8 +12,6 @@ export default function limit(limit) {
       count++
       if (count <= limit) {
         yield chunk
-      } else {
-        break
       }
     }
   }
--- a/packages/base/limit.js
+++ b/packages/base/limit.js
@@ -12,6 +12,8 @@ export default function limit(limit) {
       count++
       if (count <= limit) {
         yield chunk
+      } else {
+        break
       }
     }
   }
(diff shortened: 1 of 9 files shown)


test/e2e/pipeline.e2e.test.js
```diff
@@ -76,7 +76,7 @@ describe('Pipeline', () => {
     const out = await fromStream(env.dataset(), env.formats.parsers.import('text/turtle', pipeline.stream))
 
     // then
-    const source = await fromStream(env.dataset(), fromFile('definitions/file-loader.ttl'))
+    const source = await fromStream(env.dataset(), fromFile(env, 'definitions/file-loader.ttl'))
```
giacomociti, line 79: ah, I had noticed the missing env but I forgot to investigate

giacomociti (approved):

packages/base/limit.js
```diff
@@ -12,6 +12,8 @@ export default function limit(limit) {
       count++
       if (count <= limit) {
         yield chunk
+      } else {
+        break
```
giacomociti: I don't know if it's feasible, but we could try to inject some conditional logic at startup, so that the `break` statement is executed only if there is a next step in the pipeline.
Anyway, I would attempt this in a separate PR, I think we can merge this for now

tpluscode: That's an intriguing idea. Right now a step does not know whether it's last or not but sounds like something we can do

Alternatively, maybe a dummy `PassThrough` added ad the end of the pipeline would also solve this without any side effects?

giacomociti: still, these are all workarounds

giacomociti (approved):

PR #336, opened 2025-08-08
Update Opentelemetry

tpluscode:


ludovicm67:
What you did looks good, but the traces were not exported, and there were some errors reported, like `You can only call end() on a span once`.

I've done some changes, based on this PR.
My changes are here and are ready for review: #339 (it is targeting this PR so that you can review the changes).

ludovicm67 (approved):

PR #339, opened 2025-08-26
OpenTelemetry upgrade

ludovicm67:
The current implementation is not able to send metrics and traces properly.
This PR tries to fix those issues.


ludovicm67:
With the current changes, I'm able to export traces:

image


But I still get some issues with `You can only call end() on a span once`.

ludovicm67:
The `You can only call end() on a span once` error should be fixed now.

I also did the following:
- added a new line at the end after writing on stdout, as else the results were hidden while running it in the shell or it would require to pipe it to `cat` which is not the best in terms of user experience
- added a try…catch in the shutdown part, to avoid infinite loop in case it was not able to send metrics at the end of a pipeline
- if `otlp` is not defined, I'm setting the env vars to be set to `none` as else the NodeSDK would default to OpenTelemetry for all cases

About the metrics, it was able to export some about the memory heap, …
But I didn't have any long pipeline that I can quickly try for seeing more metrics.

I did my tests using:

```sh
OTEL_EXPORTER_OTLP_ENDPOINT=<otel-endpoint> npx barnard59 --otel-debug=VERBOSE --otel-metrics-exporter=otlp --otel-traces-exporter=otlp run test/e2e/definitions/world-clock/async.ttl
```

Could you give it a try?

PR #340, opened 2025-08-26
[CI] Use Node LTS (and LTS-1) instead of hard-coded values + use cache

ludovicm67:
Right now, the workflows are referring to Node 18 and 20, but the current LTS is not covered.
For better maintenance, it could be great to have them automatically updated, so that we don't have to maintain the version manually here.
It is also great to ensure that everything is still able to run as expected on the newer LTS versions.

Using the cache is helpful to save some precious time on the different workflow run.

### DIFF
--- a/.changeset/rich-symbols-grab.md
+++ b/.changeset/rich-symbols-grab.md
@@ -1,0 +1,5 @@
+---
+"barnard59": patch
+---
+
+Bump `approvals` to 7.3.0.
--- a/.changeset/rich-symbols-grab.md
+++ b/.changeset/rich-symbols-grab.md
@@ -1,5 +1,0 @@
----
-"barnard59": patch
----
-
-Bump `approvals` to 7.3.0.
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -7,8 +7,8 @@ jobs:
       fail-fast: false
       matrix:
         node:
-          - "18"
-          - "20"
+          - "lts/*"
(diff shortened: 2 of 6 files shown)


.github/workflows/ci.yaml
```diff
@@ -92,26 +99,26 @@ jobs:
...
     strategy:
       fail-fast: false
       matrix:
-        env:
-          - os: ubuntu-latest
-            node: 20
-          - os: macos-latest
-            node: 20
-          - os: windows-latest
-            node: 20.12.1 # Pinned, because of https://github.com/approvals/Approvals.NodeJS/issues/176
```
ludovicm67, line 108: According to https://github.com/approvals/Approvals.NodeJS/pull/177#issuecomment-2842624499, it might be fixed in a newer version 🤔

ludovicm67: I had to upgrade it to 7.3.0 so that it is not failing anymore

.changeset/rich-symbols-grab.md
```diff
@@ -0,0 +1,5 @@
+---
+"barnard59": patch
+---
+
+Bump `approvals` to 7.3.0.
```
tpluscode: It's a dev dependency, so I don't think we really care :)

ludovicm67: Good point, I will remove the changelog then 👍

ludovicm67: Done!

PR #350, opened 2026-03-27
N3

giacomociti:
added a package for N3 rules, implemented with the [eyeling](https://github.com/eyereasoner/eyeling) library


.changeset/empty-parents-return.md
```diff
@@ -0,0 +1,5 @@
+---
+"barnard59-n3-rules": minor
```
tpluscode: First of all, not super important, by why not just `-n3`?

packages/n3-rules/pipeline/applyRules.ttl
```diff
@@ -0,0 +1,53 @@
...
+      rdfs:label "Set to 'all' to include the input dataset" ;
+      p:required false ;
+    ] ;
+  ] ;
+  p:steps
+    [
+      p:stepList
+        (
+          [ <base/stdin> () ]
+          <https://barnard59.zazuko.com/pipeline/n3-rules/_apply>
```
tpluscode: Maybe this could be a blank node?

.changeset/empty-parents-return.md
tpluscode: Ah, maybe you intended the CLI usage `b59 n3-rules apply --rules` to be sentence-like?

I think that actually supports my case and `b59 n3 apply --rules` has less repetition (and typing)

giacomociti: I would also prefer simply `n3`, I avoided it because we already have `n3` as a [format](https://github.com/zazuko/barnard59/blob/master/packages/formats/n3.js). If you think there's no risk of confusion, we can rename the package

packages/n3-rules/pipeline/applyRules.ttl
giacomociti: initially it was, but there is a test calling directly this `_apply` pipeline (I made it like in the shacl package)

.changeset/empty-parents-return.md
tpluscode: I see. I don't think this should be source for any confusion TBH

tpluscode (approved):

